PR #1028, opened 2023-06-14
codeql setup

trajan0x:
**Description**

Adds codeql scanning


trajan0x:
Currently blocked on this, can't revert this: 

image

Until [this](https://github.com/synapsecns/sanguine/actions/runs/5266275965/jobs/9519997248) completes, but we will be able to tell if it fails at the right step (upload-sarif-results)

PR #1029, opened 2023-06-14
fix ui-preview worfklow when matrix vector contains no values

trajan0x:
**Description**

Fixes ui preview vector when output contains no values

image


trajan0x:
image
 
Fix worked, merging

PR #1039, opened 2023-06-14
Removing dependency from core to ethergo since ethergo depends on core

joecroninallen:
Removing dependency from core to ethergo since ehtergo depends on core

### DIFF
--- a/core/testsuite/suite.go
+++ b/core/testsuite/suite.go
@@ -4,11 +4,10 @@ package testsuite
 import (
 	"context"
 	"fmt"
-	"github.com/ethereum/go-ethereum/params"
 	"github.com/ipfs/go-log"
 	"github.com/stretchr/testify/suite"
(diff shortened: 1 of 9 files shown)


CryptoMaxPlanck (approved):
+1

core/testsuite/suite.go
```diff
@@ -213,12 +212,5 @@ func (s *TestSuite) MustMarshall(v any) []byte {
 
 // MakeBackend creates a new backend. These backends are modified to accept the higher gas limit required by the summit harness.
 func (s *TestSuite) MakeBackend(ctx context.Context, chainID *big.Int) backends.SimulatedTestBackend {
```
trajan0x, line 214: Remove altlogether, no more circular dependency

trajan0x (changes requested):

core/testsuite/suite.go
joecroninallen: done

PR #1040, opened 2023-06-14
FE: Hotfix - Stake page balances

aureliusbtc:
Fixes stake page balances not displaying, due to a bug caused from poolId being 0 & `!stakingPoolId` returning true.

### DIFF
--- a/packages/synapse-interface/pages/stake/StakeCard.tsx
+++ b/packages/synapse-interface/pages/stake/StakeCard.tsx
@@ -2,6 +2,7 @@ import { useState, useEffect } from 'react'
 import { useAccount, useNetwork } from 'wagmi'
 import { formatUnits } from '@ethersproject/units'
 import { Zero } from '@ethersproject/constants'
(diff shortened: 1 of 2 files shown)


packages/synapse-interface/pages/stake/StakeCard.tsx
```diff
@@ -59,8 +60,8 @@ const StakeCard = ({ address, chainId, pool }: StakeCardProps) => {
   })
 
   useEffect(() => {
-    if (!address || !chainId || !stakingPoolId) return
-    getStakedBalance(`0x${address.slice(2)}`, chainId, stakingPoolId)
+    if (!address || !chainId || stakingPoolId === null) return
```
trajan0x, line 63: nice fix

@nautsimon  @bigboydiamonds  can we please be more explicit w/ these type checks going forward? If zero chain ids existed, we could easily make the same mistake there.

trajan0x (changes requested):

PR #1044, opened 2023-06-16
Feat/synapse UI token ordering

bigboydiamonds:
**Description**
Implemented updated Token Ordering System based on the following tech spec document: https://docs.google.com/document/d/12LlhATvX2EE2LD5UpF7uKyseVr15bbP_Y0FWHtBRHtQ/edit

Require 
- checks on `priorityRank` assigned to Tokens in constants file (https://github.com/synapsecns/sanguine/blob/feat/synapse-ui-chain-ordering/packages/synapse-interface/constants/tokens/master.tsx) and (https://github.com/synapsecns/sanguine/blob/feat/synapse-ui-chain-ordering/packages/synapse-interface/constants/tokens/swapMaster.ts) 

- next steps on how priorityRank should be applied to `constants/tokens/poolMaster.tsx` and `constants/tokens/sushiMaster.tsx`

0ec24cb9af29950acc5096f6bbed716f7cd734b3: [synapse-interface preview link ](https://sanguine-synapse-interface-9r8z8avo6-synapsecns.vercel.app)
0f7346b68aee4e98afb2dbc55bcc0803641e9269: [synapse-interface preview link ](https://sanguine-synapse-interface-4a7rmopyt-synapsecns.vercel.app)
e1cf307006e2c5fd68ece6b36bf77771069b3758: [synapse-interface preview link ](https://sanguine-synapse-interface-4e0ydrhyd-synapsecns.vercel.app)
39edb2dbce373c0b79bfa2708279996b65894796: [synapse-interface preview link ](https://sanguine-synapse-interface-hszgoutr5-synapsecns.vercel.app)


lawsonkight:
Hardcoding the priority rank feels brittle to me, especially if we expand to support thousands of coins. Looking at the code I can't tell why a token has one rank vs another.

For example, `MIM` has `priorityRank: 6`, which does not appear in the spec. Why is that?

Would it make more sense to dynamically rank the tokens based on their properties?

Example: 
```
priorityRank = // higher numer = higher priority
  (isOrigin && token.usdValueHeld)
  + token.symbol  == 'ETH'
  + token.symbol  == 'SYN'
  + token.gas     == true
  + token.stable  == true
  + token.partner == true
  - token.symbol  == 'nETH'
  - token.symbol  == 'nUSD'
```

This allows the reasoning behind a given priorityRank to be clear in code, automatically ranks new tokens based on their priorities, and allows for easier mass changes down the road or category weighting down the road

bigboydiamonds:
> Hardcoding the priority rank feels brittle to me, especially if we expand to support thousands of coins. Looking at the code I can't tell why a token has one rank vs another.
> 
> For example, `MIM` has `priorityRank: 6`, which does not appear in the spec. Why is that?
> 
> Would it make more sense to dynamically rank the tokens based on their properties?
> 
> Example:
> 
> ```
> priorityRank = // higher numbers have higher priority
>   (isOrigin && token.usdValueHeld)
>   + token.symbol  == 'ETH'
>   + token.symbol  == 'SYN'
>   + token.stable  == true
>   + token.partner == true
>   + isGasToken(token.symbol, chainID) == true
>   - token.symbol  == 'nETH'
>   - token.symbol  == 'nUSD'
> ```
> 
> This allows the reasoning behind a given priorityRank to be clear in code, automatically ranks new tokens based on their priorities, and allows for easier mass changes down the road or category weighting down the road

It's definitely possible for us to update our sorting method to utilize a few different attributes to determine the ordering! So far we're basing it off of `priorityRank` and then alphabetically. The method you're describing could be somewhat easily implemented once we have a good idea of what attributes will update +/- priorityRank points (if we're creating priorityRank based on a few other values, and maybe pass in an initial value)

aureliusbtc (approved):

aureliusbtc (approved):

PR #1055, opened 2023-06-25
disable confirm checks on scribe

trajan0x:
**Description**

These don't work and intermittently remove old logs. These should be removed until this is no longer the case.


trajan0x:
@nautsimon I'm going to merge this for now, but I have no idea how this is working w/o any modifications to the test

PR #1059, opened 2023-06-27
SDK Multiple Routers (CCTP) Integration

aureliusbtc:
Update bridge quote logic to support multiple routers, namely SynapseCCTPRouter.

- Bumped SDK version to 0.1.30
- Added CCTP_ROUTER_ADDRESS constant mapping
- Added SynapseCCTPRouter class to handle Cross-Chain Transfer Protocol routers
- Added getBridgeTokens method to fetch bridge tokens from multiple routers in parallel
- Added getOriginQueries and getDestinationQueries methods to fetch queries from SynapseRouter or SynapseCCTPRouter
- Added findBestQuery method to find the best query pair from origin and destination queries
- Added finalizeQuote method to finalize a quote by getting fee data and setting deadlines
- Updated bridgeQuote method to get quotes from both SynapseRouter and SynapseCCTPRouter and return the best one
- Updated bridge method to determine whether to use SynapseRouter or SynapseCCTPRouter based on the origin router address


packages/sdk-router/src/sdk.ts
```diff
@@ -39,141 +53,396 @@ class SynapseSDK {
...
+      }
+
+      return bestQuote;
     }
-  }
+
+
+
+
+  // public async oldBridgeQuote(
```
aureliusbtc, line 320: TODO: Remove before merge

packages/sdk-router/src/sdk.ts
```diff
@@ -39,141 +53,396 @@ class SynapseSDK {
...
+      tokenOut = handleNativeToken(tokenOut);
+      tokenIn = handleNativeToken(tokenIn);
+
+      const originRouters: (SynapseRouter | SynapseCCTPRouter)[] = [this.synapseRouters[originChainId], this.synapseCCTPRouters[originChainId]].filter(Boolean);
+      const destRouters: (SynapseRouter | SynapseCCTPRouter)[] = [this.synapseRouters[destChainId], this.synapseCCTPRouters[destChainId]].filter(Boolean);
+
+      let bestQuote: BridgeQuote | undefined;
+
+      for (let originRouter of originRouters) {
+        for (let destRouter of destRouters) {
```
aureliusbtc, line 273: This for loop needs to be re-worked to not create synchronous blocking processes, increasing latency. OriginCCTPRouter can't call DestinationSynapseRouter, correct, @ChiTimesChi?

Instead, OriginCCTPRouter should only be calling DestCCTPRouter.

packages/sdk-router/src/sdk.ts
```diff
@@ -39,141 +53,396 @@ class SynapseSDK {
...
+          console.log(destRouter.routerContract.address)
```
aureliusbtc, line 274: Remove console.logs

packages/sdk-router/src/sdk.ts
ChiTimesChi: Correct, every Router should only call their counterpart on the other chain.

packages/sdk-router/src/sdk.ts
```diff
@@ -39,141 +53,396 @@ class SynapseSDK {
...
-
-      // Throw error if no valid bridge tokens found
-      if (!bridgeTokens?.length) {
-        throw new Error('No bridge tokens found for this route')
+    destRouters: (SynapseRouter | SynapseCCTPRouter)[]
+  ): Promise<{ symbol: string; token: string }[][]> {
+
+    const bridgeTokensPromises = destRouters.map(async (destRouter) => {
+      // Check the cache first
+      let bridgeTokens = this.bridgeTokenCache[destChainId + '_' + tokenOut];
```
ChiTimesChi, line 83: The cache key should include the router name (or simply address) as well, otherwise this creates race conditions

packages/sdk-router/src/sdk.ts
```diff
@@ -39,141 +53,396 @@ class SynapseSDK {
...
+
+      // Get fee data
+      if (isCCTP) {
+        // Cast router to SynapseCCTPRouter
+        const cctpRouter = <SynapseCCTPRouter>router;
+
+        feeAmount = await cctpRouter.routerContract.calculateFeeAmount(
+          bestBridgeToken.token,
+          originQuery.minAmountOut,
+          //TODO: how to determine isSwap?
```
ChiTimesChi, line 219: Swap Adapter field in destination query defines where there will be a swap or not:
`isSwap = (destQuery.swapAdapter != address(0))`

packages/sdk-router/src/sdk.ts
```diff
@@ -39,141 +53,396 @@ class SynapseSDK {
...
+          false
+        );
+
+        const [relayerFee, minBaseFee, , maxFee] = await cctpRouter.routerContract.feeStructures(
+          bestBridgeToken.token
+        );
+
+        // return only the compatible fields
+        feeConfig = {
+          bridgeFee: relayerFee, // Assuming relayerFee is compatible with bridgeFee
```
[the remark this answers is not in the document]
ChiTimesChi: `relayerFee` is denominated in the same units, so should be 1:1 compatible with `bridgeFee`

Should be something like this
```typescript
feeConfig = {
  bridgeFee: relayerFee,
  minFee: isSwap ? minSwapFee : minBaseFee,
  maxFee: maxFee
};
```

packages/sdk-router/src/sdk.ts
aureliusbtc: Resolved in https://github.com/synapsecns/sanguine/pull/1059/commits/59f46fb19a1f6249b3ba41632239f55f0fbf8135

packages/sdk-router/src/sdk.ts
```diff
@@ -182,17 +457,53 @@ class SynapseSDK {
...
+        routerAdapter: originQuery.swapAdapter,
+      };
+      delete bridgeOriginQuery.swapAdapter;
+
+      bridgeDestQuery = {
+        ...destQuery,
+        routerAdapter: destQuery.swapAdapter,
+      };
+      delete bridgeDestQuery.swapAdapter;
+    }
```
ChiTimesChi, line 493: This feels weird. Is there a way to avoid this by unifying `swapAdapter` and `routerAdapter` fields? These mean exactly the same thing, reason for different naming is that CCTP Router is using the updated RouterV2 naming conventions.

aureliusbtc: This was due to Typescript and Ethers Type blocking the transaction going through. 

I reworked the types fully and unionized them across both CCTP router and synapse router. 

See this: https://github.com/synapsecns/sanguine/pull/1059/files#diff-87c018c5e6200d98efe1ba3e088f0a50c7f2afd4f1b98d0417055de255cf751eR513

Not sure of a better way to handle it at the moment, but I standardized all query getters to respond with the relevant swapAdapter or routerAdapter key depending on if CCTP or not. Should mark as a todo to write some tests for that specifically.

packages/sdk-router/src/sdk.ts
```diff
@@ -10,170 +10,563 @@ import {
...
+    )
+
+    // Iterate through each array of bridge tokens
+    for (const bridgeTokens of bridgeTokensArray) {
+      try {
+        // Fetching queries from origin router
+        const originQueries = await this.getOriginQueries(
+          originRouter,
+          tokenIn,
+          [bridgeTokens.symbol],
```
ChiTimesChi, line 501: Why do we iterate through array of bridge tokens, instead of using the array of their symbols here? Passing a full array instead of `[bridgeTokens.symbol]` should limit the amount of RPC calls to 1 no matter the amount of possible bridge symbols.

PR #1051, opened 2023-06-19
V3 Testnet

CryptoMaxPlanck:
**Description**
This PR is a workspace for v3 testnet building and redeploying. It makes changes to the agents to get a working GCP build from a goreleaser build.

PR #1068, opened 2023-06-28
Refactor Agents CLI/Config

CryptoMaxPlanck:
**Description**
The command line for the agents has many fields. This PR reduces the amount of command line args and moves them into the config, as well as streamlines a lot of configuration elements.


CryptoMaxPlanck:
Need to merge #1051, test with a new helmfile deploy in the helmfile repo, then to add the new configs into the helm charts in this repo.

joecroninallen (approved):

agents/agents/agentsintegration/agentsintegration_test.go
```diff
@@ -5,9 +5,9 @@ import (
 	"github.com/brianvoe/gofakeit/v6"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/synapsecns/sanguine/agents/agents/executor"
-	executorCfg "github.com/synapsecns/sanguine/agents/agents/executor/config"
 	"github.com/synapsecns/sanguine/agents/agents/guard"
 	"github.com/synapsecns/sanguine/agents/agents/notary"
+	executor2 "github.com/synapsecns/sanguine/agents/config/executor"
```
trajan0x, line 10: can we get rid of these "2" suffixes, if it conflicts w/ a var var should prob be renamed. If it's another package (the case here, just do executorConfig for clarity)

agents/config/executor/db.go
```diff
@@ -0,0 +1,23 @@
...
+type DBConfig struct {
+	// Type is the type of database. This can be either "sqlite" or "mysql".
+	Type string `yaml:"type"`
+	// Source is the source of the database. This can be either a path to a sqlite database or a mysql database url.
+	Source string `yaml:"source"`
+}
+
+// IsValid asserts the database connection is valid.
+func (d *DBConfig) IsValid(_ context.Context) (ok bool, err error) {
+	if d.Type != "sqlite" && d.Type != "mysql" {
```
trajan0x, line 18: use dbcommon consts here

trajan0x (changes requested):

agents/agents/agentsintegration/agentsintegration_test.go
CryptoMaxPlanck: done

agents/config/executor/db.go
CryptoMaxPlanck: done

trajan0x (approved):

Issue #874, opened 2023-05-01
Otel Scribe Metrics


trajan0x:
https://pkg.go.dev/go.opentelemetry.io/otel/metric#NewAddConfig

PR #1086, opened 2023-06-29
Monitoring tools (Otel Metrics)

nautsimon:
**Description**
Adding Otel Metric monitoring tool (histogram and counter)
https://github.com/synapsecns/sanguine/issues/874


core/metrics/meter.go
```diff
@@ -0,0 +1,70 @@
...
+	stdout "go.opentelemetry.io/otel/exporters/stdout/stdoutmetric"
+	"go.opentelemetry.io/otel/metric"
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/resource"
+	semconv "go.opentelemetry.io/otel/semconv/v1.12.0"
+	"time"
+)
+
+// InitMeter creates and sets a global meter provider.
+func InitMeter(serviceName string, interval time.Duration) error {
```
trajan0x, line 15: was going to say, this should probably be per metrics provider

core/metrics/meter.go
```diff
@@ -0,0 +1,70 @@
...
+			sdkmetric.NewPeriodicReader(exporter, sdkmetric.WithInterval(interval)),
+		),
+	)
+	otel.SetMeterProvider(mp)
+	return nil
+}
+
+// NewCounter creates a new meter counter instrument.
+// https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#counter
+func NewCounter(meterName string, counterName string, desc string, units string) (metric.Int64Counter, error) {
```
trajan0x, line 39: tbh, figure out which of these you need first and start integrating them into scribe or wherever you want and then we'll start figuring this out

trajan0x (changes requested):

core/metrics/meter.go
nautsimon: what do you think about
1. Use counters for reorg events (by chain)
2. Use a histogram for frequency of last block indexed (by chain, and use contract address as an attr). that way we can do max, min, and count on the blocks stored (max will tell us the max block height of that interval)

services/scribe/node/scribe.go
```diff
@@ -56,7 +63,11 @@ func (s Scribe) Start(ctx context.Context) error {
 	for i := range s.config.Chains {
 		chainConfig := s.config.Chains[i]
 		chainID := chainConfig.ChainID
-
+		reorgMeter, err := metrics.NewCounter(s.handler.Meter(), fmt.Sprint("scribe_reorg_counter_", chainID), "block_histogram", "a reorg meter", "reorg events")
+		if err != nil {
+			return fmt.Errorf("error creating otel counter %w", err)
+		}
+		s.reorgMeters[chainID] = reorgMeter
```
nautsimon, line 70: Creation

services/scribe/backfill/contract.go
```diff
@@ -55,7 +58,7 @@ const invalidTxVRSError = "invalid transaction v, r, s values"
 const txNotFoundError = "not found"
 
 // NewContractBackfiller creates a new backfiller for a contract.
-func NewContractBackfiller(chainConfig config.ChainConfig, contractConfig config.ContractConfig, eventDB db.EventDB, client []ScribeBackend, handler metrics.Handler) (*ContractBackfiller, error) {
+func NewContractBackfiller(chainConfig config.ChainConfig, contractConfig config.ContractConfig, eventDB db.EventDB, client []ScribeBackend, handler metrics.Handler, blockMeter otelMetrics.Int64Histogram) (*ContractBackfiller, error) {
```
nautsimon, line 61: Create backfiller with histogram meter

services/scribe/backfill/contract.go
```diff
@@ -184,6 +192,9 @@ func (c *ContractBackfiller) Backfill(parentCtx context.Context, givenStart uint
 	if err != nil {
 		return fmt.Errorf("could not store last indexed block: %w", err)
 	}
+	c.blockMeter.Record(ctx, int64(endHeight), otelMetrics.WithAttributeSet(
+		attribute.NewSet(attribute.String("contract_address", c.contractConfig.Address), attribute.Int64("chain_id", int64(c.chainConfig.ChainID)))),
+	)
```
nautsimon, line 197: log block store

core/metrics/base.go
```diff
@@ -102,11 +111,22 @@ func newBaseHandlerWithTracerProvider(buildInfo config.BuildInfo, tracerProvider
 	tracer := tracerProvider.Tracer(buildInfo.Name())
 	otel.SetTextMapPropagator(propagator)
 
+	interval, err := strconv.Atoi(os.Getenv("OTEL_METER_INTERVAL"))
+	if err != nil {
+		// default interval
+		interval = 60
+	}
```
nautsimon, line 118: gets env var, uses 60 seconds as default

core/metrics/metrics.go
```diff
@@ -31,6 +32,8 @@ type Handler interface {
 	Propagator() propagation.TextMapPropagator
 	// Type returns the handler type.
 	Type() HandlerType
+	// Meter returns an otel meter provider
+	Meter() *sdkmetric.MeterProvider
```
trajan0x, line 36: should return an interface so you can no-op things. You basically a need a way to have the functions people need to call from Meter() without panicing, but nothing gets passed into them

core/metrics/null.go
```diff
@@ -50,6 +52,9 @@ func (n nullHandler) Gin() gin.HandlerFunc {
 func (n nullHandler) Start(_ context.Context) error {
 	return nil
 }
+func (n nullHandler) Meter() *sdkmetric.MeterProvider {
+	return nil
```
trajan0x, line 56: this isn't gonna work, this'll nil dereference here the second you try to use it:

image

core/metrics/null.go
```diff
@@ -75,4 +80,8 @@ func (n nullPropogator) Fields() []string {
 	return []string{}
 }
 
+func (n nullPropogator) Meter() *sdkmetric.MeterProvider {
```
trajan0x, line 83: Meter's don't need to be propgated since they're not tracked cross service

core/metrics/meter_test.go
```diff
@@ -0,0 +1,47 @@
...
+		attribute.NewSet(attribute.Int64("block_number", 32), attribute.Int64("chain_id", 1))),
+	)
+}
+
+// Recording values in a histogram format, such the actual block numbers stored over time.
+func TestHistogram(t *testing.T) {
+	mp, err := metrics.InitMeter("test", time.Duration(30)*time.Second)
+	Nil(t, err)
+
+	histogram, err := metrics.NewHistogram(mp, "test", "block_histogram", "a block counter", "blocks")
```
trajan0x, line 36: what are your thoughts on this pattern vs mp.NewHistogram()

core/metrics/meter.go
```diff
@@ -0,0 +1,67 @@
...
+	"go.opentelemetry.io/otel/sdk/resource"
+	semconv "go.opentelemetry.io/otel/semconv/v1.12.0"
+	"time"
+)
+
+// InitMeter creates and sets a global meter provider.
+func InitMeter(serviceName string, interval time.Duration) (*sdkmetric.MeterProvider, error) {
+	// TODO configure exporter how we need.
+
+	exporter, err := stdout.New()
```
trajan0x, line 17: make this passable

services/scribe/backfill/contract.go
trajan0x: do we really nee to pass this in again?

services/scribe/backfill/contract_test.go
```diff
@@ -60,7 +62,10 @@ func (b BackfillSuite) TestFailedStore() {
 		StoreConcurrency:   1,
 		GetLogsRange:       1,
 	}
-	backfiller, err := backfill.NewContractBackfiller(chainConfig, contractConfig, mockDB, simulatedChainArr, b.metrics)
+	blockHeightMeter, err := metrics.NewHistogram(b.metrics.Meter(), fmt.Sprint("scribe_block_meter", chainConfig.ChainID), "block_histogram", "a block height meter", "blocks")
```
trajan0x, line 65: yeah, the point of the nil provider is you don't have to pass these in in the tests

trajan0x (changes requested):

core/metrics/meter.go
trajan0x: like this to start actually. Let's get one of these meters up and then iterate from there

core/metrics/meter.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// MeterImpl is an implementation of the MeterProvider interface.
+type MeterImpl struct {
+	mp *sdkmetric.MeterProvider
+}
+
+// NewMeter creates a new meter provider.
+func NewMeter(serviceName string, interval time.Duration, exporter sdkmetric.Exporter) (*MeterImpl, error) {
+	resource := resource.NewWithAttributes(
```
trajan0x, line 27: change the name here since this shadows an import

trajan0x (changes requested):

PR #1087, opened 2023-06-29
Add `RelaySingle` command to CCTP relayer

dwasse:
**Description**
Adds a new `relay-single` command that allows processing a single tx via CLI.

Given an `origin` and `tx-hash`, the relayer will fetch the corresponding attestation and submit the `ReceiveCircleToken` transaction on the destination chain. The command will wait until the message is marked as `Complete` via the `CircleRequestFulfilled` event.


trajan0x (changes requested):
Think this needs some work since the resubmission queue will stop before this has submitted and the tx might not go through

PR #1091, opened 2023-06-29
CCTP relayer listens for `CircleRequestFulfilled` events

dwasse:
**Description**
`CCTPRelayer` will now listen for `CircleRequestFulfilled` events and store corresponding messages with `Complete` state. If a corresponding pending message does not exist in the db, we reconstruct what we can from the event log.

- Add `GetMessageByRequestID` db accessor
- Move getters to `CCTPRelayerDBReader` interface
- Add `DestTxHash` to relayer `Message` along with corresponding db field


services/cctp-relayer/relayer/relayer.go
```diff
@@ -593,9 +673,10 @@ func (c *CCTPRelayer) submitReceiveCircleToken(parentCtx context.Context, msg *r
 	}
 
 	// Store the completed message.
-	// Note: this can cause double sbumissin sometimes
+	// Note: this can cause double submission sometimes
 	msg.State = relayTypes.Submitted
 	msg.DestNonce = int(nonce)
+	msg.DestTxHash = destTxHash.String()
```
trajan0x, line 679: This won't work accurately because of the way submitter works rn, but we can deal w/ this later

trajan0x (changes requested):

services/cctp-relayer/relayer/relayer.go
trajan0x: This should be ready w/ the merging of #1098

services/cctp-relayer/db/sql/base/message.go
```diff
@@ -120,3 +123,19 @@ func (s Store) GetMessageByOriginHash(ctx context.Context, originHash common.Has
 
 	return &message, nil
 }
+
+// GetMessageByRequestID gets a message by its request id.
+// TODO: this is actually non-unique, but we only return 1.
```
trajan0x, line 128: Not true

trajan0x (changes requested):

PR #1092, opened 2023-06-30
[DO NOT MERGE] test out a google trace provider

trajan0x:
**Description**

**Additional context**

**Metadata**
- Fixes #[Link to Issue]


trajan0x:
After a lot of experimentation, I think this is going to be way too tedious to setup since it requires setting up a service account for each namespace, binding that service account to the kube service account an then allowing workload identity assumption

PR #1104, opened 2023-07-03
Feat/synapse UI portfolio

bigboydiamonds:
**Description**
Adding a `Portfolio` component that will display an overview of a connected wallet's balances on all supported networks / tokens. Portfolios will display a connect wallet's tokens in a approved / unapproved => highest -> lowest held parsed balance format. 

Portfolio component consists of the following major components: 
- `PortfolioTabManager` that keeps track of what Portfolio section is currently active / shown
- `SingleNetworkPortfolio` component that houses all the information around a single network (network name, id, filtered held tokens based on approved/unapproved + parsed balances in descending order) as well as user actions such as connecting to another network
- `PortfolioAccordion` that is in charge of expanded/collapsed state of a `SinglePortfolioNetwork`

Integrated Portfolio into the existing Redux state-managed Bridge page.


packages/synapse-interface/styles/global.css
```diff
@@ -195,3 +195,12 @@ body {
 .no-mt {
   margin-top: 0 !important;
 }
+
+.radial-gradient-underline {
```
abtestingalpha, line 199: fy we can do this with tailwind + inline for consistency - example from landing page:

```
 <Button
  className={`
    border-[#AC8FFF] border text-sm
    px-4 py-3 hover:opacity-75
  `}
  style={{
    background: 'linear-gradient(310.65deg, rgba(255, 0, 255, 0.2) -17.9%, rgba(172, 143, 255, 0.2) 86.48%)',
    borderRadius: '10px',
  }}>
    Go to Explorer
</Button>
```

bigboydiamonds: great point - thanks for pointing that out! I'm leaning towards keeping it as a custom global class (need to read into tailwind's docs to see how I can set this natively within the tailwind system) just so it's easier to grab this specific styling for multiple buttons. I'm thinking these styles/colors will likely be applied across multiple components, so wanted to prevent having to find all the instances before updating

abtestingalpha: For the app, we don't want to clutter up the global css file bc it will start to get tough to manage. Would prefer to use it if we have something super custom (like the loader animations, which we can't do in tw). There's some stuff in there for the landing page as well rn, which we should strip out too (all can be done in tw, and aren't repeated elsewhere either). We're also, atm, not re-using this css elsewhere and if it comes to the point we are, then we'll have a `<SpecialButton />` component instead.

Solution is just inline:

```
//removed radial-gradient-bg up here and added as inline style

const ConnectedButton = () => {
  const buttonClassName = `
  h-9 flex items-center justify-center w-36
  text-base text-white px-4 py-2 rounded-3xl
  text-center transform-gpu transition-all duration-75
  border-2 border-[#D747FF]
  hover:cursor-default
  `

  return (
    <button
      data-test-id="connected-button"
      className={buttonClassName}
      style={{
        background:
          'radial-gradient(249.66% 351.84% at -119.61% 97.91%, #FF00FF 0%, #AC8FFF 100%), linear-gradient(90deg, rgba(128, 0, 255, 0.2) 0%, rgba(255, 0, 191, 0.2) 100%)',
      }}
    >
      Connected
    </button>
  )
}
```

for `radial-gradient-underline` in `PortfolioTabManager`

```
style={{
  borderImage: 'linear-gradient(to right, rgba(255, 0, 255, 1), rgba(172, 143, 255, 1)) 1'
}}
```

packages/synapse-interface/pages/state-managed-bridge/index.tsx
```diff
@@ -149,6 +151,12 @@ const StateManagedBridge = () => {
   const fromChainIds = Object.keys(CHAINS_BY_ID).map((id) => Number(id))
   const toChainIds = Object.keys(CHAINS_BY_ID).map((id) => Number(id))
 
+  useEffect(() => {
```
abtestingalpha, line 154: Need to remove this useEffect, breaking ability to switch chains

packages/synapse-interface/pages/state-managed-bridge/index.tsx
```diff
@@ -460,9 +468,10 @@ const StateManagedBridge = () => {
         data-test-id="bridge-page"
         className="relative z-0 flex-1 h-full overflow-y-auto focus:outline-none"
       >
-        <div className="items-center px-4 py-20 mx-auto mt-4 2xl:w-3/4 sm:mt-6 sm:px-8 md:px-12">
-          <div className="flex flex-col items-center justify-center">
-            <div className="flex items-center space-x-20">
+        <div className="flex flex-col justify-center px-4 py-20 mx-auto mt-4 lg:flex-row 2xl:w-3/4 sm:mt-6 sm:px-8 md:px-12">
+          <Portfolio />
```
abtestingalpha, line 472: See comment in 

`packages/synapse-interface/pages/index.tsx`

packages/synapse-interface/styles/global.css
bigboydiamonds: Updated in https://github.com/synapsecns/sanguine/pull/1104/commits/2cdfc16cba1cd60ddce9be3bb51d6f1a2abb95b0

aureliusbtc (changes requested):

packages/synapse-interface/utils/hooks/usePortfolioBalances.tsx
```diff
@@ -1,86 +1,154 @@
...
-  chainId: number
-) => {
-  const [balances, setBalances] = useState([])
+  const { chain } = useNetwork()
+  const { address } = getAccount()
+  const availableChains = Object.keys(BRIDGABLE_TOKENS)
+  const filteredChains = availableChains.filter((chain) => chain !== '2000') // need to figure out whats wrong with Dogechain
 
-  let calls = []
+  const fetchPortfolioBalances = async () => {
```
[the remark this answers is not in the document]
bigboydiamonds: Thank you! Added this in commit https://github.com/synapsecns/sanguine/pull/1104/commits/9254e3e56240aae72e9078c2c336e8ace0b31148

packages/synapse-interface/components/Portfolio/Portfolio.tsx
```diff
@@ -0,0 +1,96 @@
...
+            fetchState={fetchState}
+            bridgeTxHashes={bridgeTxHashes}
+          />
+        )}
+      </div>
+    </div>
+  )
+}
+
+function filterPortfolioBalancesWithBalances(
```
aureliusbtc, line 80: The filterPortfolioBalancesWithBalances function could be improved by using .filter() instead of .forEach() and re-creating the object. This would be more efficient:

```
function filterPortfolioBalancesWithBalances(balancesAndAllowances) {
  return Object.fromEntries(
    Object.entries(balancesAndAllowances).map(([key, tokenWithBalances]) => {
      const filtered = tokenWithBalances.filter(t => t.balance > Zero)  
      return [key, filtered]
    })
  )
}
```

packages/synapse-interface/components/Portfolio/PortfolioContent.tsx
```diff
@@ -0,0 +1,176 @@
...
+        <React.Fragment>
+          <PortfolioAssetHeader />
+          <UnconnectedPortfolioContent />
+        </React.Fragment>
+      )}
+    </div>
+  )
+}
+
+function getCurrentNetworkPortfolio(
```
aureliusbtc, line 90: The getCurrentNetworkPortfolio helper could be improved by using filter instead of deleting:

```
function getCurrentNetworkPortfolio(currentChainId, networks) {

  const currentNetwork = networks[currentChainId]  

  const remainingNetworks = Object.keys(networks)
    .filter(key => key !== currentChainId)
    .map(key => ({[key]: networks[key]}))

  return { currentNetwork, remainingNetworks }

}
```

aureliusbtc (approved):

packages/synapse-interface/components/Portfolio/Portfolio.tsx
bigboydiamonds: That's a great point! Using `.filter()` method would be more efficient than the latter. 

Made a slight adjustment to the code you provided - needed to add an additional `.filter()` call to filter out Portfolios that didn't hold balances (otherwise all networks would display, regardless of balance): 

`
function filterPortfolioBalancesWithBalances(
  balancesAndAllowances: NetworkTokenBalancesAndAllowances
): NetworkTokenBalancesAndAllowances {
  const filteredEntries = Object.entries(balancesAndAllowances).map(
    ([key, tokenWithBalances]) => {
      const filtered = tokenWithBalances.filter(
        (token: TokenWithBalanceAndAllowance) => token.balance > Zero
      )
      return [key, filtered]
    }
  )

  const filteredEntriesWithValues = filteredEntries.filter(
    ([_, filtered]) => filtered.length > 0
  )

  return Object.fromEntries(filteredEntriesWithValues)
}
`
Updated in commit https://github.com/synapsecns/sanguine/pull/1104/commits/260ff5296f20b56f65a9f989586177ecaa94cc8b

bigboydiamonds: Further optimizing function by using `.reduce()` method as opposed to `.map() + .filter()`: 

```
function filterPortfolioBalancesWithBalances(
  balancesAndAllowances: NetworkTokenBalancesAndAllowances
): NetworkTokenBalancesAndAllowances {
  return Object.entries(balancesAndAllowances).reduce(
    (
      filteredBalances: NetworkTokenBalancesAndAllowances,
      [key, tokenWithBalances]
    ) => {
      const filtered = tokenWithBalances.filter(
        (token: TokenWithBalanceAndAllowance) => token.balance > Zero
      )
      if (filtered.length > 0) {
        filteredBalances[key] = filtered
      }
      return filteredBalances
    },
    {}
  )
}
```

Updated in commit https://github.com/synapsecns/sanguine/pull/1104/commits/deef779f25ed6ad26245c01379f07ace728c10d7

packages/synapse-interface/components/Portfolio/PortfolioContent.tsx
bigboydiamonds: In order to satisfy the conditions of `currentNetwork` and `remainingNetworks` being objects with key:array key-value pairs (in order to properly pass network:token data to the Portfolio component) , we would have to re-create the object as such: 

```
function getCurrentNetworkPortfolio(
  currentChainId: number,
  networks: NetworkTokenBalancesAndAllowances
): {
  currentNetwork: NetworkTokenBalancesAndAllowances
  remainingNetworks: NetworkTokenBalancesAndAllowances
} {
  const currentNetwork: NetworkTokenBalancesAndAllowances = {
    [currentChainId]: networks[currentChainId],
  }

  const remainingNetworks: NetworkTokenBalancesAndAllowances = Object.keys(
    networks
  )
    .filter((key: string) => Number(key) !== currentChainId)
    .reduce((result: NetworkTokenBalancesAndAllowances, key: string) => {
      result[Number(key)] = networks[key]
      return result
    }, {})

  return { currentNetwork, remainingNetworks }
}
```

Given the additional overhead using methods to iterate of keys of an object, think it would be more efficient to instead delete the key-value pair directly from the `remainingNetworks` object.

PR #1107, opened 2023-07-05
Add CCTP event indexing to explorer

dwasse:
**Description**
Adds the necessary log handling for CCTP events in explorer:
- `CircleRequestSent`
- `CircleRequestFulfilled`

**Additional context**
- Add a couple helper utils for parsing Nullable sql values


services/explorer/types/cctp/event.go
```diff
@@ -0,0 +1,47 @@
+package cctp
+
+import (
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// EventLog is the interface for all bridge events.
```
nautsimon, line 9: for all cctp events.

services/explorer/db/sql/model.go
```diff
@@ -77,6 +77,50 @@ var (
...
+	// RequestID is the request ID of the CCTP transfer.
+	RequestID [32]byte `gorm:"column:request_id"`
+	// Recipient is the recipient of the CCTP transfer.
+	Recipient sql.NullString `gorm:"column:recipient"`
+	// Fee is the fee of the CCTP transfer.
+	Fee *big.Int `gorm:"column:fee;type:UInt256"`
+	// FeeUSD is the fee of the CCTP transfer in USD terms.
+	FeeUSD *float64 `gorm:"column:fee_usd;type:Float64"`
+	// Token is the address of the received token.
+	Token sql.NullString `gorm:"column:token"`
```
nautsimon, line 119: what will we be using this column for if we have burn and mint

also are we gonna use/cool with using fee as the difference between origin/destination?

nautsimon: another idea is:
- have a non-pointer based column for chain id. put the chain where the event occurred in there
- have destination or origin populated depending on the event

nautsimon: i think using fee is good too though

services/explorer/consumer/parser/cctpparser.go
```diff
@@ -0,0 +1,134 @@
...
+	timeStamp, err := c.consumerFetcher.FetchBlockTime(ctx, int(chainID), int(iFace.GetBlockNumber()))
+	if err != nil {
+		return nil, fmt.Errorf("could not get block time: %w", err)
+	}
+
+	// If we have a timestamp, populate the following attributes of cctpEvent.
+	timeStampBig := uint64(*timeStamp)
+	cctpEvent.TimeStamp = &timeStampBig
+
+	err = c.applyPriceData(ctx, &cctpEvent, usdcCoinGeckoID)
```
nautsimon, line 93: 😎

nautsimon:
I think this is great, all we got to do now is add tests. will dm!

services/explorer/db/sql/model.go
dwasse: Hmm yeah probably easiest to use fee as a switch for Sent/Fulfilled events.

I think the ideal way to do this would be to have `RequestID` be the primary key and handle OnConflict clauses to only update non-null fields if a record already exists for the given `RequestID`. Not sure where this logic would go though (also seems to break the pattern that you use in other event processing).

Another possibility would be to have separate models for Sent/Fulfilled events and leave it up to the db reader to join on `RequestID`, but again, depends on explorer access patterns and would require more boilerplate.

nautsimon (approved):
tested locally, lgtm

PR #1108, opened 2023-07-05
Updated scribe helm chart

bsgrigorov:
Set resources per deployment for scribe


nautsimon:
Will confirm values tomorrow when we check in

trajan0x (approved):

PR #1120, opened 2023-07-10
Fix: allow API to return results with decimals

dwasse:
**Description**
Allows the API to return results that have decimal values.

**Additional context**
For example, a query to `/bridge?fromChain=1&toChain=42161&fromToken=gOHM&toToken=gOHM&amount=10` would result with `maxAmountOutStr` of 9, when the underlying human-readable value is actually 9.995. After applying this fix we get the expected result of 9.995.


nautsimon (approved):
lgtm, tested locally

Defi-Moses:
lgtm

PR #1167, opened 2023-07-26
Feat/rest api upgrade

Defi-Moses:
**Description**
This upgrade reconfigures dependencies so that it is compatible with heroku deployment, adds two new functions that allow users to generate the raw transaction data in python (so they can acutually bridge once connected to Web3), and updates the ReadMe. 

**Additional context**
Ideally @nautsimon @dwasse can review the two new functions (very slight changes) to confirm that the output transaction data is correct. This also probably needs more robust testing
**Metadata**
- Fixes #[Link to Issue]

### DIFF
--- a/packages/rest-api/app.ts
+++ b/packages/rest-api/app.ts
@@ -2,12 +2,37 @@ import { JsonRpcProvider } from '@ethersproject/providers'
 import { SynapseSDK } from '@synapsecns/sdk-router'
 import { BigNumber } from '@ethersproject/bignumber'
 import { formatUnits } from '@ethersproject/units'
-import express from 'express'
+import * as express from 'express'
+// import express from 'express'
 
-import chains from './config/chains.json'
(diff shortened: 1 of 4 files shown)


packages/rest-api/app.ts
```diff
@@ -220,6 +249,169 @@ app.get('/bridge', async (req, res) => {
...
+    !fromTokenAddress ||
+    !toTokenAddress ||
+    !fromTokenDecimals ||
+    !toTokenDecimals
+  ) {
+    res.send(
+      `
+      <h1>Invalid Params</h1>
+      <hr/>
+      <b>Ensure that your request matches the following format: /swap?chain=1&fromToken=USDC&toToken=DAI&amount=100</b>
```
dwasse, line 286: Probably want `/swapTxInfo` instead of `/swap` here right?

dwasse:
LGTM

PR #1173, opened 2023-07-27
Explorer Bridgewatcher

nautsimon:
**Description**
Adding new endpoints to explorer for retrieving origin and destination txs for bridgewatcher to
- reduce rpc load on the fe
- provide more reliability
- build off existing services stack


trajan0x:
Since this is live/may cause metrics conflicts, merging.

Note: this doesn't fallback to rpcs yet

PR #1189, opened 2023-07-31
Added prometheus ports to omnirpc helm chart

bsgrigorov:


charts/omnirpc/templates/deployment.yaml
```diff
@@ -53,9 +53,10 @@ spec:
             {{- toYaml .Values.env | nindent 12 }}
           {{- end }}
           ports:
-            - name: http
-              containerPort: 5080
-              protocol: TCP
+          {{- range .Values.service.ports }}
```
trajan0x, line 56: do we want to export the metrics at the service level, or the container level? Not sure which we're doing here but container is definitely preferable

charts/omnirpc/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" }}
```
trajan0x, line 1: think we may want to allow users do disable?

trajan0x (changes requested):

charts/omnirpc/templates/deployment.yaml
bsgrigorov: service level is more standard pod level should be used only if we don't have a service

charts/omnirpc/templates/servicemonitor.yaml
bsgrigorov: ok

charts/omnirpc/templates/deployment.yaml
trajan0x: I don't think so. The container label exists for a reason by default on prom. This way you can do aggegates across all your containers.

If you do service level you're just hitting a random container every scrape interval and you miss any outliers in your histogram

trajan0x (changes requested):

trajan0x (approved):

trajan0x (approved):

PR #1195, opened 2023-07-31
[DO NOT MERGE] CCTP relayer context handling

dwasse:
Attempt to fix an issue in CCTP relayer where the application would hang upon context cancellation.


trajan0x:
The idea here, is, if this fixes the issue we should see multiple resets in gcp

trajan0x:
This appears to have solved the issue. The pod this is in has reset 2x, so we do have a context leak somewhere. I'm unsure where

PR #1202, opened 2023-08-01
Explorer Server updates

nautsimon:
**Description**
- [x] Added cctp event search with all standard filters (pending, chain, amount, time, etc.)
- [x] Updated partial info type, return information about event type in bridge query (for fe updates)
~Bridgewatcher fallback~ (for another PR)

Future
~- add more bespoke cctp queries depending on requests from standup~

**Status**
d483b40fcd729c29c6a018c4ca7d395032af16a1 in prod


nautsimon:
will be merged in #1248

PR #1228, opened 2023-08-02
De-duplicate agent signing logic

dwasse:
Adds `Encoder` interface to allow for de-duplication of agent type signing logic.

### DIFF
--- a/agents/types/encoder.go
+++ b/agents/types/encoder.go
@@ -20,6 +20,11 @@ const (
 	uint40Len = 5
 )
 
+// Encoder encodes a type to bytes.
+type Encoder interface {
+	Encode() ([]byte, error)
+}
+
 // EncodeGasData encodes a gasdata.
 func EncodeGasData(gasData GasData) ([]byte, error) {
 	b := make([]byte, 0)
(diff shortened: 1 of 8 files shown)


agents/types/encoder.go
CryptoMaxPlanck, line 1: thoughts on deleting this file and moving each specific encoder to its specific type file?

agents/types/utils.go
```diff
@@ -0,0 +1,35 @@
...
+	// Hash the encoded type, and concatenate with hashed salt.
+	hashedEncoded := crypto.Keccak256Hash(encoded).Bytes()
+	toSign := append(crypto.Keccak256Hash([]byte(salt)).Bytes(), hashedEncoded...)
+	hashedDigest, err := HashRawBytes(toSign)
+	if err != nil {
+		return nil, nil, common.Hash{}, fmt.Errorf("could not hash receipt: %w", err)
+	}
+
+	// Sign the message.
+	signature, err := signer.SignMessage(ctx, core.BytesToSlice(hashedDigest), false)
```
CryptoMaxPlanck, line 30: is there ever a scenario where we would want to sign a payload hash without the `\x19Ethereum...`?

CryptoMaxPlanck (approved):
looks great! really cleans things up and makes the code much more legible

agents/types/utils.go
CryptoMaxPlanck: ah actually i dont think so. ignore.

agents/types/encoder.go
dwasse: Yeah I think eventually that's the right approach, if we like the approach I can go ahead and implement for all the other types (although we still have to define the `Encoder` interface somewhere).

There's a question of how to implement the decoder functions, I think ideally we could do something like this:

```go
type Encoder interface {
  Encode() ([]byte, error)
  Decode([]byte) error
}
```

Where the new pattern to decode would be:

```go
receipt := new(Receipt)
err = receipt.Decode(rawBytes)
```

agents/types/utils.go
dwasse: Not that I know of! If we need to customize this can always extend this function or add a new one

PR #1231, opened 2023-08-04
Executor messages to summit

CryptoMaxPlanck:
**Description**
Adds support for executor messages going to summit. Both base and manager.


agents/agents/executor/types/contract_type.go
```diff
@@ -14,6 +14,8 @@ const (
 	LightInboxContract // LightInboxContract
 	// InboxContract is a TODO.
 	InboxContract // InboxContract
+	// SummitContract is a TODO.
```
trajan0x, line 17: Why is this a todo?

trajan0x (changes requested):
Looks pretty good.

One thing I'm curious about is if anecdotally you guys have been seeing a higher % of errors after switching to anvil backend

agents/agents/executor/types/contract_type.go
CryptoMaxPlanck: :p, let me fix these

CryptoMaxPlanck:
and @trajan0x the error rate has been pretty consistent (which is not good, but not worse)

PR #1235, opened 2023-08-04
Fix:sdk same chain ids

ChiTimesChi:
- `bridge()` and `bridgeQuote()` functions now throw an error, if equal chain ids were passed for origin and destination chains.
- Added test coverage.

### DIFF
--- a/packages/sdk-router/src/sdk.test.ts
+++ b/packages/sdk-router/src/sdk.test.ts
@@ -252,6 +252,22 @@ describe('SynapseSDK', () => {
       expect(data?.length).toBeGreaterThan(0)
       expect(to?.length).toBeGreaterThan(0)
     })
+    it('throws if origin == destination', async () => {
+      const chainIds = [43114]
+      const providers = [avalancheProvider]
+      const Synapse = new SynapseSDK(chainIds, providers)
+      await expect(
+        Synapse.bridgeQuote(
+          43114,
+          43114,
+          '0xCFc37A6AB183dd4aED08C204D1c2773c0b1BDf46',
+          '0xA7D7079b0FEaD91F3e65f86E8915Cb59c1a4C664',
+          BigNumber.from('1000000000000000000000')
+        )
+      ).rejects.toThrowError(
+        'Origin chainId cannot be equal to destination chainId'
+      )
+    })
   })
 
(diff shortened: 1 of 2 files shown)


packages/sdk-router/src/sdk.test.ts
```diff
@@ -409,6 +425,36 @@ describe('SynapseSDK', () => {
...
+          1,
+          '',
+          BigNumber.from('10000000000000000000'),
+          originQuery!,
+          destQuery!
+        )
+      ).rejects.toThrowError(
+        'Origin chainId cannot be equal to destination chainId'
+      )
+    })
```
abtestingalpha, line 457: This looks good. Think if we wanted to refactor repeated uses of a successful bridgeQuote down the line, we could use `beforeEach` and a `setupTest` helper

```
describe('bridge', () => {
  let Synapse
  let result
  let originQuery
  let destQuery

  const setupTest = async (fromChainId, toChainId, tokenIn, tokenOut, amountIn, chainIds, providers) => {
    Synapse = new SynapseSDK(chainIds, providers);
    result = await Synapse.bridgeQuote(
      fromChainId,
      toChainId,
      tokenIn,
      tokenOut,
      amountIn
    );
    if (!result) {
      throw Error
    }
    originQuery = result.originQuery;
    destQuery = result.destQuery;
  }

  describe('valid chain ids', () => {
    beforeEach(async () => {
      await setupTest(testVars)
    });

    it('successful bridge', async () => {
     const { data, to } = await Synapse.bridge(
       ...vars,
       originQuery!,
       destQuery!
     )
     expect(data?.length).toBeGreaterThan(0)
     expect(to?.length).toBeGreaterThan(0)
    })
  })

  describe('invalid chain ids', () => {
    beforeEach(async () => {
      await setupTest(testVars)
    });

    it('chain ids are the same, throws error', async () => {
      await expect(Synapse.bridge(...vars, originQuery!, destQuery!)).rejects.toThrowError('Origin chainId cannot be equal to destination chainId')
    })
  
    it('chain ids we dont support', async () => {
      // throws different error
    })
  })
})
```

abtestingalpha (approved):

trajan0x:
LGTM

PR #1239, opened 2023-08-08
Add RPC head check + [goreleaser]

nautsimon:
**Description**
Handle when unreliable RPCs (eg. boba, canto) return a head block hundreds or thousands of blocks behind the true head.


nautsimon:
Will be merged in #1248

PR #1274, opened 2023-08-22
Devnet

trajan0x:
**Description**

You can test this out from `packages/contracts-core` with the following command:

`make devnet-clean && make devnet-up && make devnet-deploy`

PR #1280, opened 2023-08-24
Fraud reporting

dwasse:
**Description**

Adds fraud reporting functionality to the guard.

Handles [these cases](https://gist.github.com/ChiTimesChi/906aa1d56b0c2914838d809d81ab2385) as well as updating agent statuses.:

| Fraud type                        | Agent        | Fraud proving                | Fraud reporting                    | Description                                  |
| --------------------------------- | ------------ | ---------------------------- | ---------------------------------- | -------------------------------------------- |
| Invalid Snapshot                  | Notary/Guard | verifyStateWithSnapshot      | submitStateReportWithSnapshot (N)  | Snapshot having State not matching Origin    |
| Invalid Attestation               | Notary       | verifyAttestation            | submitAttestationReport            | Attestation that was never created in Summit |
| Attestation from Invalid Snapshot | Notary       | verifyStateWithAttestation   | submitStateReportWithAttestation   | Attestation having State not matching Origin |
| Attestation from Invalid Snapshot | Notary       | verifyStateWithSnapshotProof | submitStateReportWithSnapshotProof | (alternative way to prove/report fraud)      |
| Invalid Receipt                   | Notary       | verifyReceipt                | submitReceiptReport                | Receipt that doesn't match Destination       |


agents/agents/guard/db/sql/base/base.go
```diff
@@ -33,6 +33,7 @@ func (s Store) SubmitterDB() submitterDB.Service {
 // GetAllModels gets all models to migrate.
 // see: https://medium.com/@SaifAbid/slice-interfaces-8c78f8b6345d for an explanation of why we can't do this at initialization time
 func GetAllModels() (allModels []interface{}) {
+	allModels = append(allModels, &Dispute{}, &AgentTree{}, &AgentRoot{})
```
trajan0x, line 36: make this one line w/ 37

agents/agents/guard/db/sql/base/model.go
```diff
@@ -0,0 +1,71 @@
...
+
+// AgentTree is the state of an agent tree on Summit.
+type AgentTree struct {
+	// AgentRoot is the root of the agent tree.
+	AgentRoot string `gorm:"column:agent_root;primaryKey"`
+	// AgentAddress is the address of the agent for the Merkle proof.
+	AgentAddress string `gorm:"column:agent_address;primaryKey"`
+	// BlockNumber is the block number that the agent tree was updated (on summit).
+	BlockNumber uint64 `gorm:"column:block_number"`
+	// TODO: Check compatibility with [][32]byte.
```
trajan0x, line 60: done?

agents/agents/guard/guard.go
```diff
@@ -27,59 +41,225 @@ import (
...
+	// Set up evm utilities for each domain
 	for domainName, domain := range cfg.Domains {
-		var domainClient domains.DomainClient
+		omnirpcClient, err := omniRPCClient.GetConfirmationsClient(ctx, int(domain.DomainID), 1)
+		if err != nil {
+			return nil, fmt.Errorf("error with omniRPCClient, could not create guard: %w", err)
+		}
 
 		chainRPCURL := omniRPCClient.GetEndpoint(int(domain.DomainID), 1)
-		domainClient, err = evm.NewEVM(ctx, domainName, domain, chainRPCURL)
```
trajan0x, line 63: line above should be GetDefaultEndpoint

agents/agents/guard/fraud.go
```diff
@@ -0,0 +1,504 @@
...
+		if !shouldSubmit {
+			return nil
+		}
+
+		// Submit the state report.
+		srSignature, _, _, err := state.SignState(ctx, g.bondedSigner)
+		if err != nil {
+			return fmt.Errorf("could not sign state: %w", err)
+		}
+		_, err = g.domains[g.summitDomainID].Inbox().SubmitStateReportWithSnapshot(
```
trajan0x, line 57: fyi, this will deref if g.summitDomainID set incorrectly

agents/agents/guard/guard.go
```diff
@@ -232,6 +414,37 @@ func (g Guard) Start(parentCtx context.Context) error {
 		return err
 	})
 
+	group.Go(func() error {
```
trajan0x, line 417: group->g, let's keep this consistent

agents/agents/guard/suite_test.go
```diff
@@ -23,10 +23,10 @@ func NewGuardSuite(tb testing.TB) *GuardSuite {
 	}
 }
 
-func (u *GuardSuite) SetupTest() {
+func (g *GuardSuite) SetupTest() {
 	chainwatcher.PollInterval = time.Second
 
-	u.SimulatedBackendsTestSuite.SetupTest()
+	g.SimulatedBackendsTestSuite.SetupTest()
```
trajan0x, line 29: remove the previous line (chainwatcher.PollInterval)

agents/domains/evm/destination.go
```diff
@@ -49,6 +49,8 @@ func (a destinationContract) Execute(transactor *bind.TransactOpts, message type
 		return nil, fmt.Errorf("could not encode message: %w", err)
 	}
 
+	a.nonceManager.ClearNonce(transactor.From)
```
trajan0x, line 52: why?

agents/domains/evm/inbox.go
```diff
@@ -42,12 +45,47 @@ type inboxContract struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("could not encode signature: %w", err)
+	}
+
+	// TODO: Is there a way to get a return value from a contractTransactor call?
+	tx, err = a.contract.SubmitStateReportWithSnapshot(transactOpts, big.NewInt(stateIndex), rawSig, snapPayload, snapSignature)
+	if err != nil {
+		// TODO: Why is this done? And if it is necessary, we should functionalize it.
+		if strings.Contains(err.Error(), "nonce too low") {
+			a.nonceManager.ClearNonce(signer.Address())
```
trajan0x, line 74: again, why?

agents/domains/evm/inbox.go
```diff
@@ -58,3 +96,181 @@ func (a inboxContract) SubmitSnapshot(transactor *bind.TransactOpts, signer sign
...
+	rawSig, err := types.EncodeSignature(signature)
+	if err != nil {
+		return nil, fmt.Errorf("could not encode signature: %w", err)
+	}
+
+	// TODO: Is there a way to get a return value from a contractTransactor call?
+	tx, err = a.contract.SubmitStateReportWithAttestation(transactOpts, big.NewInt(stateIndex), rawSig, snapPayload, attPayload, attSignature)
+	if err != nil {
+		// TODO: Why is this done? And if it is necessary, we should functionalize it.
+		if strings.Contains(err.Error(), "nonce too low") {
```
trajan0x, line 189: why?

agents/domains/evm/inbox.go
```diff
@@ -58,3 +96,181 @@ func (a inboxContract) SubmitSnapshot(transactor *bind.TransactOpts, signer sign
...
+
+	transactOpts.GasLimit = 5000000
+
+	a.nonceManager.ClearNonce(signer.Address())
+
+	// TODO: Is there a way to get a return value from a contractTransactor call?
+	tx, err = a.contract.SubmitReceiptReport(transactOpts, rcptPayload, rcptSignature, rrSignature)
+	if err != nil {
+		// TODO: Why is this done? And if it is necessary, we should functionalize it.
+		if strings.Contains(err.Error(), "nonce too low") {
```
trajan0x, line 269: why?

agents/domains/evm/lightinbox.go
```diff
@@ -61,3 +63,127 @@ func (a lightInboxContract) SubmitAttestation(
...
+		return nil, fmt.Errorf("could not sign tx: %w", err)
+	}
+
+	transactOpts, err := a.nonceManager.NewKeyedTransactor(transactor)
+	if err != nil {
+		return nil, fmt.Errorf("could not create tx: %w", err)
+	}
+
+	transactOpts.Context = ctx
+	transactOpts.GasLimit = 5000000
```
trajan0x, line 170: hardcoding these is a bad idea. This is going to be too low for l2s and too high for other chains

agents/types/chain_gas.go
```diff
@@ -35,4 +37,19 @@ func (g chainGas) Domain() uint32 {
 	return g.domain
 }
 
+// ChainGassesToSnapGas converts a slice of ChainGas to a slice of big.Int.
+func ChainGassesToSnapGas(chainGasses []ChainGas) (snapGasses []*big.Int, err error) {
+	snapGasses = make([]*big.Int, len(chainGasses))
+	for i, chainGas := range chainGasses {
```
trajan0x, line 43: rename this var, it's shadowed by the struct

agents/types/dispute_processed_status.go
```diff
@@ -0,0 +1,14 @@
+package types
+
+// DisputeProcessedStatus is the status of a dispute on Summit.
+// This enum is used for tracking the status of disputes in the Dispute db model.
+type DisputeProcessedStatus uint8
```
trajan0x, line 5: shouldn't this be a db.Enum?

agents/types/dispute_status.go
```diff
@@ -0,0 +1,54 @@
...
+func (s disputeStatus) Rival() common.Address {
+	return s.rival
+}
+
+func (s disputeStatus) FraudProver() common.Address {
+	return s.fraudProver
+}
+
+func (s disputeStatus) DisputePtr() *big.Int {
+	return s.disputePtr
```
trajan0x, line 51: core.CopyBigInt. *ANY* Caller that mutates this will change the underlying struct without this safety mechanism

trajan0x (changes requested):

agents/agents/guard/cmd/commands.go
```diff
@@ -122,7 +124,22 @@ var GuardRunCommand = &cli.Command{
 
 			g, _ := errgroup.WithContext(c.Context)
 
-			guard, err := guard.NewGuard(c.Context, guardConfig, baseOmniRPCClient, guardDB, handler)
+			embedded := client.NewEmbeddedScribe(
```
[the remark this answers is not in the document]
CryptoMaxPlanck: Fixed this to add different scribe support.

agents/agents/guard/db/sql/base/base.go
CryptoMaxPlanck: Don't think there is a way to do this in one `append`, but can make it in one line by stringing together two `append`s. Going to keep separate for now.

agents/agents/guard/fraud.go
CryptoMaxPlanck: Now checking for this in constructor, so we know we will have it.

agents/agents/guard/guard.go
CryptoMaxPlanck: Receiver name here is already `g`. Going to keep it as `group`, unless you think we should either a) change the receiver, b) change from `group` to something else (non `g`)

agents/types/dispute_processed_status.go
CryptoMaxPlanck: What is the purpose of doing this? Seems like overhead. Does it help with querying in some way?

trajan0x:
Re: flakes

I've noticed we've got a lot of very, very busy tests running at once:

image


Agentintegration is insanely over generalized and shouldn't be a base case, but as long as it is, there's no way we should be parallelizing this on just 2 cores. Especailly w/ the anvil cpu overhead.

Erego, going to try this:

```go
#!/bin/bash
# Initialize coverage.txt
echo "mode: atomic" > coverage.txt

# Loop through all packages and run tests
for pkg in $(go list ./...); do
    go test -coverpkg="$pkg" -coverprofile=profile.cov $pkg
    if [ -f profile.cov ]; then
        tail -n +2 profile.cov >> coverage.txt
        rm profile.cov
    fi
done
```

If it works, I'm gonna need to re-enable pyroscope somehow

trajan0x:
FYI guys im testing this out on devnet rn, at a high level code stuff looks good.  Some todos on my end:

- Reenable pyroscope. Test script in use now is an mvp from here https://github.com/synapsecns/sanguine/pull/1280#issuecomment-1704401113 and needs a tiny bit of work
- Implement https://github.com/golang/go/issues/15513 Should substantially bring down test time

agents/domains/evm/lightinbox.go
```diff
@@ -61,3 +62,70 @@ func (a lightInboxContract) SubmitAttestation(
...
 }
+
+//nolint:dupl
+func (a lightInboxContract) SubmitStateReportWithSnapshot(transactor *bind.TransactOpts, stateIndex int64, signature signer.Signature, snapPayload []byte, snapSignature []byte) (tx *ethTypes.Transaction, err error) {
+	rawSig, err := types.EncodeSignature(signature)
+	if err != nil {
+		return nil, fmt.Errorf("could not encode signature: %w", err)
+	}
+
+	// TODO: Is there a way to get a return value from a contractTransactor call?
```
trajan0x, line 73: Not something we have to fix for this pr, but I am exploring a few soltuions here:

https://twitter.com/trajan0x/status/1701008059991232897

agents/config/agent_config.go
```diff
@@ -17,6 +18,8 @@ import (
 type AgentConfig struct {
 	// DBConfig is the database configuration.
 	DBConfig DBConfig `yaml:"db_config"`
+	// ScribeConfig is the scribe configuration.
+	ScribeConfig ScribeConfig `yaml:"scribe_config"`
```
trajan0x, line 22: Not something to address in this pr, but as @aureliusbtc pointed out I think the entire process of setting up all these configs is extremely confusing when not using an embedded scribe. tbh, I think I pushed a bit too much for flexibility over canonicality when speccing out embedded vs remote scribe. 

Anyway, we can address this in #1274, I just wanted to note this here

agents/contracts/inbox/parser.go
```diff
@@ -67,6 +79,8 @@ type EventType uint
 const (
 	// SnapshotAcceptedEvent is a SnapshotAccepted event.
 	SnapshotAcceptedEvent EventType = iota
+	// ReceiptAcceptedEvent is a ReceiptAccepted event.
+	ReceiptAcceptedEvent
```
trajan0x, line 83: iota + 1 everywhere

agents/contracts/inbox/topics.go
```diff
@@ -19,20 +19,30 @@ func init() {
 
 	SnapshotAcceptedTopic = parsedInbox.Events["SnapshotAccepted"].ID
 
+	ReceiptAcceptedTopic = parsedInbox.Events["ReceiptAccepted"].ID
+
 	if SnapshotAcceptedTopic == (common.Hash{}) {
 		panic("SnapshotAcceptedTopic is nil")
 	}
+	if ReceiptAcceptedTopic == (common.Hash{}) {
```
trajan0x, line 27: nice

agents/contracts/inbox/parser.go
trajan0x: btw, we should note in release notes or something this will break existing dbs

agents/testutil/deployers.go
```diff
@@ -21,6 +21,9 @@ import (
 	"github.com/synapsecns/sanguine/ethergo/deployer"
 )
 
+// SynChainID the id of the SynChain.
+const SynChainID uint = 10
```
trajan0x, line 25: fyi, not sure if we want to change this here or someone else but we should at least note as of #1308, this no longer needs to be hardcoded (a change we made for both #1274 and #1312)

agents/testutil/simulated_backends_suite.go
```diff
@@ -148,7 +152,8 @@ func (a *SimulatedBackendsTestSuite) SetupSuite() {
 	a.TestSuite.LogDir = filet.TmpDir(a.T(), "")
 
 	// don't use metrics on ci for integration tests
-	useMetrics := core.GetEnvBool("CI", true)
+	isCI := core.GetEnvBool("CI", false)
```
trajan0x, line 155: Does this work?

ethergo/backends/anvil/anvil.go
```diff
@@ -121,7 +122,7 @@ func NewAnvilBackend(ctx context.Context, t *testing.T, args *OptionBuilder) *Ba
 	// to prevent old containers from piling up, we set a timeout to remove the container.
 	require.Nil(t, resource.Expire(args.expirySeconds))
 
-	address := fmt.Sprintf("%s:%s", "http://localhost", resource.GetPort("8545/tcp"))
+	address := fmt.Sprintf("%s:%s", "http://localhost", dockerutil.GetPort(resource, "8545/tcp"))
```
trajan0x, line 125: Okay, I'm not sure how relevant this is, but I want to mention it because of all the work you guys did getting this ready in #1330 and #1334 is to uplaod the docker logs everywhere we use [TailContainerLogs](https://pkg.go.dev/github.com/synapsecns/sanguine/core/dockerutil#TailContainerLogs)

Especially in [anvil](https://pkg.go.dev/github.com/synapsecns/sanguine/ethergo/backends/anvil) + core[https://pkg.go.dev/github.com/synapsecns/sanguine/core/metrics/localmetrics].

We could then export these as workflow artifacts like we did in 1290 here:

https://github.com/synapsecns/sanguine/pull/1290/files#diff-38442ca3eaca86339f50860b9eaa0b38301559f536d66a37d7200a23c117821eR235-R246

image
 
and they'd be downoadable like [this](https://github.com/synapsecns/sanguine/actions/runs/6038167381):

image

(As an aside, we could also do this w/ all the jaeger tracing stuff possibly, that's just a lot hell of a lot more work, considering we're moving in the direction of disabling jaeger in tests, this could really just be an otlp file)

This way, when flakes do occur, they'd be much easier to track down.

If this makes sense, would appreciate it if guys could make a new issue and just reference this otherwise curious to here why it wouldn't be helpful

core/dockerutil/port_util.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"github.com/ory/dockertest/v3"
+	dc "github.com/ory/dockertest/v3/docker"
+)
+
+// GetPort returns the port of a container.
+// unlike resource.GetPort this returns the ports in reverse order.
+// this can be useful in ci environments(?)
+func GetPort(resource *dockertest.Resource, port string) string {
```
trajan0x, line 11: Let's add a bit more explanation here. Tl;dr is before we were returning a port on 0.0.0.0 something docker does not force to be unique, now we just return the last port to get around the 0.0.0.0 thing on non docker-for-mac stuff (you guys have as much context as I do on why this works.  Just ran docker ps and noticed ports were different sometimes, perhaps chatgpt can explain?) if not just explain the behavior  int the comment and say we're not sure why it works

trajan0x: Side note. Can we please implement this for explorer/localmetrics in this PR? Just anywhere we're using resource.GetPort:

image


I have a feeling this is also causing explorer flakes (Cc @nautsimon )

ethergo/backends/base/base.go
```diff
@@ -150,8 +151,17 @@ func (b *Backend) VerifyContract(contractType contracts.ContractType, contract c
...
-			assert.NotEmpty(b.T(), code)
+			require.Nil(b.T(), err)
+			if len(code) == 0 {
+				receipt, err := b.Client().TransactionReceipt(b.ctx, contract.DeployTx().Hash())
+				require.Nil(b.T(), err)
+
+				jsonReceipt, err := receipt.MarshalJSON()
+				require.Nil(b.T(), err)
+
+				fmt.Println(string(jsonReceipt))
```
trajan0x, line 162: delete me, this was all temporary stuff for debugging

ethergo/chain/chain.go
```diff
@@ -39,6 +39,8 @@ type Chain interface {
 	Estimator() gas.PriceEstimator
 	// GasSetter gets the gas setter
 	GasSetter() gas.Setter
+	// ChainConfig gets the chain config.
+	ChainConfig() *params.ChainConfig
```
trajan0x, line 43: if you want you can remove this, but probably good to have. As an aside, we've really gotta deprecate this entire chain package soon.

ethergo/submitter/suite_test.go
```diff
@@ -91,8 +92,16 @@ func (s *SubmitterSuite) SetupSuite() {
...
 		defer wg.Done()
 		var err error
-		localmetrics.SetupTestJaeger(s.GetSuiteContext(), s.T())
-		s.metrics, err = metrics.NewByType(s.GetSuiteContext(), buildInfo, metrics.Jaeger)
+		// don't use metrics on ci for integration tests
+		isCI := core.GetEnvBool("CI", false)
+		useMetrics := !isCI
+		metricsHandler := metrics.Null
+
+		if useMetrics {
```
trajan0x, line 100: does this actually work?

trajan0x (changes requested):
Guys, most of this looks good.But I have been going through trying to understand what's happening at a high level, especially making sure we didn't break anything in CI since this applies to a lot of go modules.

Do that end, I setup a quick checklist here and made some changes. https://github.com/synapsecns/sanguine/pull/1339#issue-1888770174

One of my most significant discoveries was that we're currently not exiting if status is 1 from our for loop (full explanation, courtesy of chatgpt here: https://github.com/synapsecns/sanguine/pull/1339/files/0c88d79a8e86375f56455d289760c35fb7aa18c2#r1320848293)

So, in the last workflow on this pr the [agents workflow](https://github.com/synapsecns/sanguine/actions/runs/6125555460/job/16628246448) is still passing, but [ethergo workflow](https://github.com/synapsecns/sanguine/actions/runs/6125555460/job/16628247414) is failing, something that wasn't before/historically hasn't even flaked. 

We should get this figured out. Once we do I think we'll be good to revert https://github.com/synapsecns/sanguine/pull/1339/commits/913228c61b32f99ab9c27711f6086f81e8255660 and then we should be good to GTM. 

I know this whole flake thing has been a pain, but  comprehensive tests will pay dividends for every contributor going forward.

I've closed #1330 and #1334 as these seemed supplemental to this pr and now that i's ready unneccesray.

One thing I've focused on in my review is a few comments not specific to this pr, but around process considering how long it took to iron out these flakes. Other than more comprehensive code reviews earlier on would be curious what kinda tooling would make these flakes easier to iron out. I've added my suggestions.

One more aside,  it's not a major point but can we figure out what's reordering import/doing stuff like this and disable it or find a way to make our lint workflow force standardization? 

image

the reason I say, other than it makes reading diffs/reviewing revision histories more annoying is it also just causes a lot of ci stuff to run on things that haven't really changed. Again, minor thing and doesn't have to be done for this pr, just a future note

Finally, do you guys think you could add a general explanation of the testing suite somewhere? @ChiTimesChi  does this for solidity [here](https://github.com/synapsecns/sanguine/tree/master/packages/contracts-core/test#testing-suite-for-the-messaging-contracts) and I think it's really helpful for anyone coming into it w/ fresh eyes or that hasn't looked in a while.

Might also be good to add a little readme to the guard that contains [these cases](https://gist.github.com/ChiTimesChi/906aa1d56b0c2914838d809d81ab2385) considering how complex this whole thing is

ethergo/backends/anvil/anvil.go
trajan0x: FYI, we could also directly embed this into https://pkg.go.dev/github.com/synapsecns/sanguine/core@v0.0.77/testsuite#TestSuite.TearDownSuite, but this would be a lot quicker to just do once at the end of the workflow. My reco:

1. Specify a `ARTIFACT_DIR`
2. Write all thse filets tehre
3. Zip + upload w/ ${matrix.package}/docker-logs.zip.

Note: while their exist actions like [this](https://github.com/marketplace/actions/collect-docker-logs) these won't work because we cleanup docker logs since we are [resource constrained](https://github.com/synapsecns/sanguine/pull/1280#issuecomment-1704401113)

trajan0x: Update: implemented a thing here: https://github.com/synapsecns/sanguine/pull/1339/files#diff-678682767f2477de3e3c546746f8568b9a1942b2c647d32331d7e774b8ff8d9fR204-R227

agents/agents/guard/fraud.go
```diff
@@ -0,0 +1,792 @@
...
+			}
+
+			return
+		})
+		if err != nil {
+			return fmt.Errorf("could not submit VerifyStateWithSnapshot tx: %w", err)
+		}
+
+		// Check if we should submit the state report.
+		shouldSubmit, err := g.shouldSubmitStateReport(ctx, fraudSnapshot)
```
ChiTimesChi, line 55: I feel like should submit logic should be applied to both Synapse Chain and remote chain. For instance:
- Someone has reported a fraudulent Notary on SynChain, but not on remote chain. Report on remote chain only.
- Someone has reported a fraudulent Notary on remote chain, but not on SynChain. Report on SynChain only.
- No one has reported a fraudulent Notary. Report on both chains.

agents/agents/guard/fraud.go
```diff
@@ -0,0 +1,792 @@
...
+		}
+
+		return nil
+	}
+	err = retry.WithBackoff(ctx, contractCall, g.retryConfig...)
+	if err != nil {
+		return false, fmt.Errorf("could not get dispute status: %w", err)
+	}
+
+	isNotSummit := snapshot.AgentDomain != 0
```
ChiTimesChi, line 141: Shouldn't this be called `isNotGuard` / `isNotary`?

SynChain doesn't have any assigned Notaries, and Guards are assigned to every chain (they have zero value for domain to showcase that).

agents/agents/guard/fraud.go
```diff
@@ -0,0 +1,792 @@
...
+		isSlashable, err := g.isStateSlashable(ctx, state)
+		if err != nil {
+			return fmt.Errorf("could not check if state is slashable: %w", err)
+		}
+		if !isSlashable {
+			continue
+		}
+
+		// Initiate slashing on origin.
+		_, err = g.txSubmitter.SubmitTransaction(ctx, big.NewInt(int64(state.Origin())), func(transactor *bind.TransactOpts) (tx *ethTypes.Transaction, err error) {
```
ChiTimesChi, line 239: See above note about `fraudAttestation.Agent` status on `state.Origin()` chain

agents/agents/guard/fraud.go
```diff
@@ -0,0 +1,792 @@
...
+			)
+			if err != nil {
+				return nil, fmt.Errorf("could not submit state report with snapshot to summit: %w", err)
+			}
+
+			return
+		})
+		if err != nil {
+			return fmt.Errorf("could not submit SubmitStateReportWithSnapshot tx: %w", err)
+		}
```
ChiTimesChi, line 91: Might be overthinking this, but could we explore the generalization of submission of the report transaction in the future?

Something like "here's the chain Id, here's the signed report (of any type), here's some data to accompany it, please submit the report tx for me".

ChiTimesChi: Outlining this, as the core logic is pretty much the same:

https://github.com/synapsecns/sanguine/blob/f5eb4de20c67c5a570cdab50139334a51197e5e3/agents/agents/guard/fraud.go#L101-L117

https://github.com/synapsecns/sanguine/blob/f5eb4de20c67c5a570cdab50139334a51197e5e3/agents/agents/guard/fraud.go#L270-L287

etc

agents/agents/guard/fraud.go
```diff
@@ -0,0 +1,792 @@
...
+			if err != nil {
+				return fmt.Errorf("could not get agent status: %w", err)
+			}
+
+			return nil
+		}
+		err = retry.WithBackoff(ctx, contractCall, g.retryConfig...)
+		if err != nil {
+			return false, fmt.Errorf("could not get agent status: %w", err)
+		}
```
ChiTimesChi, line 339: Similarly, could this be later further abstracted as `DomainClient.GetAgentStatus()`?

agents/contracts/bondingmanager/topics.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+func init() {
+	// set topics
+	var err error
+
+	parsedBondingManager, err := abi.JSON(strings.NewReader(BondingManagerMetaData.ABI))
```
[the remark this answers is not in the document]
dwasse: The pattern is bearable, but yeah if there's a way to autogenerate all topics that would be great. IMO it would be less error-prone, and also potentially helpful for debugging where we can parse topics on a whim to inspect auxiliary events.

dwasse:
> One more aside, it's not a major point but can we figure out what's reordering import/doing stuff like this and disable it or find a way to make our lint workflow force standardization?

Regarding imports, my recommendation would be we add a precommit hook that runs something like this:

```bash
find . -name "*.go" -exec goimports -w {} \;
```

That way regardless of editor choice we will have `gofmt`-styled code as well as standardized import ordering (`goimports` is a superset of `gofmt` that includes import organization).

cc @trajan0x

ethergo/submitter/suite_test.go
dwasse: We observed it working when running `watch -n 1 'docker ps'` manually in upterm

agents/agents/guard/fraud.go
dwasse: Makes sense- opted for `isNotary`

core/dockerutil/port_util.go
dwasse: >Side note. Can we please implement this for explorer/localmetrics in this PR? Just anywhere we're using resource.GetPort

Addressed [here](https://github.com/synapsecns/sanguine/pull/1280/commits/3a1dcb84d39e121ae8ea0bab90af56ff399026f8)

dwasse: > Let's add a bit more explanation here.

Addressed [here](https://github.com/synapsecns/sanguine/pull/1280/commits/bbd958f2f3e39e9b81fc8857569fc248c766dda9)

agents/testutil/deployers.go
dwasse: Added a TODO

agents/testutil/simulated_backends_suite.go
dwasse: Responded below!

PR #1287, opened 2023-08-28
Extra state reports

dwasse:
After a `Guard` submits a state report (either with snapshot or with attestation), we first submit the report on Summit, then on the agent's respective domain. If the agent is not yet registered on the remote domain, we store a `RelayableAgentStatus` model and trigger a call to `updateAgentStatus()` on the domain prior to submitting the report.

The `RelayableAgentStatus` model is mean to replace the `Dispute` model for the sake of generality; it allows us to handle both the case where `updateAgentStatus()` must be called as a result of a remote dispute, or an agent that is yet to be registered.


agents/agents/guard/db/sql/base/agent_root.go
```diff
@@ -36,12 +36,11 @@ func (s Store) StoreAgentRoot(
...
-
+func (s Store) GetSummitBlockNumberForRoot(ctx context.Context, agentRoot string) (uint64, error) {
 	var blockNumber uint64
 	dbTx := s.DB().WithContext(ctx).
-		Where(fmt.Sprintf("%s = ?", AgentRootFieldName), dbAgentRoot).
+		Where(fmt.Sprintf("%s = ?", AgentRootFieldName), agentRoot).
+		Order(fmt.Sprintf("%s ASC", BlockNumberFieldName)).
 		Limit(1).
 		Model(&AgentRoot{}).
 		Pluck(BlockNumberFieldName, &blockNumber)
```
CryptoMaxPlanck, line 46: `Pluck` vs `Scan`?

agents/agents/guard/db/sql/base/crosstable.go
```diff
@@ -11,42 +11,52 @@ import (
...
-// 2. Filter the rows where the `DisputeProcessedStatus` is `Resolved`.
+type agentTreeWithStatus struct {
+	AgentTree
+	AgentDomain      uint32
+	UpdatedAgentFlag agentTypes.AgentFlagType
+}
+
+// GetRelayableAgentStatuses gets the parameters for updating the agent status with the following steps:
+// 1. Outer join the `AgentTree` table on the `RelayableAgentStatus` table on the `AgentAddress` fields.
+// 2. Filter the rows where the `AgentStatusRelayedState` is `Queued`.
```
CryptoMaxPlanck, line 22: nit: comment order doesn't match query order. 1 and 2 should be swapped.

agents/agents/guard/db/sql/base/crosstable.go
```diff
@@ -11,42 +11,52 @@ import (
...
 // 3. Return each of remaining rows' `AgentRoot`, `AgentAddress`, and `Proof` fields.
```
CryptoMaxPlanck, line 23: also this comment should be updated since we return much more now

CryptoMaxPlanck (approved):
tiny stuff here n there. other than that, lgtm!

agents/agents/guard/db/sql/base/agent_root.go
dwasse: My understanding is that `Scan` will load results into a given struct, whereas `Pluck` will do a SELECT for an individual field in the query

PR #1288, opened 2023-08-28
Submitter and Retries for Fraud

CryptoMaxPlanck:
**Description**
Utilize the submitter and retries in the guard's fraud transactions.


agents/agents/guard/fraud.go
```diff
@@ -21,6 +24,7 @@ func (g Guard) handleSnapshot(ctx context.Context, log ethTypes.Log) error {
 
 	// Verify each state in the snapshot.
 	for stateIndex, state := range fraudSnapshot.Snapshot.States() {
+		stateIndex, state := stateIndex, state
```
dwasse, line 27: This is to avoid value mirroring when we dispatch to the submitter right? Do you think it's more readable to change the loop variable names? Could go either way really, I'm just being a nit :)

Something like:

```go
	for i, s := range fraudSnapshot.Snapshot.States() {
		stateIndex, state := i, s
```

agents/agents/guard/fraud.go
```diff
@@ -101,7 +123,17 @@ func (g Guard) handleSnapshot(ctx context.Context, log ethTypes.Log) error {
 // Only submit a state report if we are not on Summit, and the snapshot
 // agent is not currently in dispute.
 func (g Guard) shouldSubmitStateReport(ctx context.Context, snapshot *types.FraudSnapshot) (bool, error) {
-	disputeStatus, err := g.domains[g.summitDomainID].BondingManager().GetDisputeStatus(ctx, snapshot.Agent)
+	var disputeStatus types.DisputeStatus
+	var err error
+	contractCall := func(ctx context.Context) error {
```
dwasse, line 128: At some point we want to allow configurable backoff logic. In the `Guard` we could do something like:

```go
g.backoffStrat = g.cfg.MaxRetryDurationSeconds
```

And we could make the contract call like this:

```go
err = retry.WithBackoff(ctx, contractCall)
```

Up to you if we do it in this PR or merge this one and impl the backoff config in the base PR!

dwasse (approved):
LGTM- a couple minor comments

agents/agents/guard/fraud.go
CryptoMaxPlanck: agreed!

agents/agents/guard/fraud.go
CryptoMaxPlanck: just added

PR #1302, opened 2023-09-01
Notary Fix Origin State

CryptoMaxPlanck:
**Description**
Make the notary use states that originate from the Notary's domain.


CryptoMaxPlanck:
Adding executor test fix via backend bump. This fixes flake of TestSendManagerMessage

PR #1308, opened 2023-09-04
set the synapse domain in the constructor

trajan0x:


ChiTimesChi:
@trajan0x Solidity changes LGTM

PR #1312, opened 2023-09-05
Deploy Testnet Contracts

aureliusbtc:
## Summary by CodeRabbit

The existing bullet-point list is still valid based on the provided instructions.


trajan0x:
LFG!

Not sure how we want to deal with the constructor stuff in 6fe7d4aacb3c0655a8c4eb974682ca4dea31714b consider testnet contracts already deployed, but I think I can just copy the deployments over to the devnet branch (#1274)

PR #1331, opened 2023-09-07
Fix: passing agent root order [messaging]

ChiTimesChi:
**Description**
Remote `Destination` contracts are tracking the "pending" `agentRoot`, which is automatically passed to the local `LightManager` contract once its optimistic period passes. This is exposed via public method `passAgentRoot()`, which is also called whenever a new attestation is submitted.

Previously, it was called prior to handling the new attestation, which resulted in agent root passing, but the attestation was silently discarded (as the Notary was not deemed active under the newly passed agent root). This is weird from the Notary's perspective who've paid gas to effectively update the agent root and now has to re-register themselves and submit **the same attestation**.

The new solution handles the attestation first, and only then attempts to pass the agent root. This establishes an easy invariant:
- Either Notary attestation is accepted and transaction succeeds.
- Or Notary attestation is not accepted and transaction reverts.

**Additional context**
- TBD: order of merging this and #1308

### Summary by CodeRabbit

- Refactor: Simplified the `passAgentRoot` function across several contracts and interfaces, which now returns a single boolean value `rootPending` instead of a tuple. This change streamlines the process of passing agent roots, improving the system's efficiency.
- New Feature: Enhanced the `Destination` contract to handle different scenarios more effectively, including checks for changes in the agent root, dispute status, and the end of the optimistic period.
- Test: Updated and added new tests in the `DestinationTest` contract to ensure the correct behavior of the updated `passAgentRoot` function and the handling of different scenarios.
- Chore: Updated the metadata of several contracts with new ABI and Bin values, ensuring the system is up-to-date with the latest changes.


trajan0x:
@chitimeschi didn't see the note here about merging this and #1308. Ended up merging that first because it's required for #1274 and (ideally) future deployments of #1312. That being said, I'll go ahead and regenerate agents for this one

trajan0x:
Updated agents in 6f89ec54d34a2e97167cf3e7ff0269c7cbffcc3e

mikeyrf (approved):
Code looks good.

PR description states "The new solution handles the attestation first, and only then attempts to pass the agent root", but not seeing where call to `passAgentRoot()` was added? Likely missing something so ignore if so.

ChiTimesChi:
@mikeyrf no **extra** calls were added, but the order of the calls has been changed:

Now:
https://github.com/synapsecns/sanguine/blob/1b1ea755b1a948707d362f6389b3f0555c372a26/packages/contracts-core/contracts/Destination.sol#L98-L101

Old:
https://github.com/synapsecns/sanguine/blob/4e12b934ba403aed10a8188603782a3552f18ca2/packages/contracts-core/contracts/Destination.sol#L99-L110

One other change is that we no longer silently exit if the agent root was passed, instead we do the Notary verification using the previous root, and then attempt to pass the new root, if there is one.

Old:
https://github.com/synapsecns/sanguine/blob/4e12b934ba403aed10a8188603782a3552f18ca2/packages/contracts-core/contracts/Destination.sol#L99-L102

aureliusbtc (approved):

ChiTimesChi:
Closing, to be replaced by:
- #1526

PR #1339, opened 2023-09-09
Fraud Report Final Touches 🎉

trajan0x:
**Description**

Tracking in: https://github.com/synapsecns/sanguine/pull/1280#issuecomment-1712525480

- [x] Implement some optimizations suggested in https://github.com/golang/go/issues/15513 (tbh, I'm not even sure if these do anything. Can re-evaluate)
- [ ] Re-enable pyroscope
- [x] Fix codecov
- [x] Get script to fail if any test fails  (see [here](https://github.com/synapsecns/sanguine/actions/runs/6138844667/job/16656124535?pr=1339). Test fails, but actin passes just w/ lower coerage

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -165,7 +165,7 @@ jobs:
         with:
           command: |
             cd ${{ matrix.package }}
-            go test -run=nope ./...  # pre-compile all the tests (w/o running)
+            GOMEMLIMIT=5Gi go test -run=nope ./...  # pre-compile all the tests (w/o running). At this point, no containers are running so we can also up the memlimit
             echo "mode: atomic" > coverage.txt
             for pkg in $(go list ./...); do
               go test -coverpkg="$pkg" -coverprofile=profile.cov $pkg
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -160,19 +160,29 @@ jobs:
...
+        env:
+          # no other containers are running at this point so we can use all available memory for compilation
+          GOMEMLIMIT: 5GiB
+          GOGC: -1
+
       - name: Test
         uses: nick-fields/retry@v2
         with:
           command: |
+            set -e
```
trajan0x, line 175: ## Explanation: Why the Script Didn't Stop on Errors

The provided script looped through each package and ran a series of commands on each package. If any command failed during this process, the script would simply continue to the next package in the list without stopping. 

This behavior is due to the default nature of bash scripting. By default, a bash script won't terminate or exit just because a command inside it fails. Instead, it will continue executing subsequent commands.

If you want a bash script to exit immediately when a command fails, you need to add the `set -e` directive at the beginning of the script. The `set -e` directive instructs the shell to exit immediately if any command in the script returns a non-zero exit status (indicating failure).

Without the `set -e` directive, your script will continue running subsequent commands even if one of them fails, which is the behavior you observed.

PR #1370, opened 2023-09-13
Removes legacy dispatched type from OriginHub

trajan0x:
**Description**

Removes now unused `Dispatch` type from `OriginHub`


trajan0x:
ignore codecov for this, this was caused by [ci skip] being in the commit body when #1280 was merged

PR #1381, opened 2023-09-15
Devnet

trajan0x:
**Note: This pr was moved from #1274 because I want @coderabbitai to participate in review and it didn't appear to be doing this for older prs **

**Description**

The following PR adds a fully featured devnet for local testing in just 2 commands. From `packages/contracts-core`:

1. `make devnet-up`
2. `make devnet-deploy`

Now you can use the devnet however you want as specified in here:

https://github.com/synapsecns/sanguine/blob/03a8b00fd8a0f0afdc32a13bacf06067076352c6/packages/contracts-core/README.md#running-a-devnet

I've tested this out a few times and have yet to have any issues, but feedback appreciated. Additionally, this makes a fair number of changes to scripts and it'd appear I wasn't as diligent about spotting those as I could've been in #1308 (see #1376). 

Among the open questions I have:

- [ ] @ChiTimesChi: Is `packages/contracts-core/create3/Create3.sol` the right place for this contract?
- [ ] @aureliusbtc: Do we want to move the testnet deployments from #1312 in here. Is there anything we're missing?
- [ ] Do we want to move `ICreate3Factory` to a config file? If yes, is this pr the right place to do that?
- [ ] Clean out old scribe params that are confusing to users: #1385 

## Future work:

- I'd very much like to setup an integration test around this. I decided not to go the golang route I was [contemplating on twitter a few weeks ago ](https://twitter.com/trajan0x/status/1694127968035979292) and looks like op decided not to either ethereum-optimism/optimism#6710 That makes testing a bit harder so I'm going to hold off for now. Build time of all these containeres is also going to be a pain
- Grafana. I have prometheus enabled, but it's really not that helpful w/o a grafana instance other than for some more obscure debugging stuff. Alerts would also be good
- Config Guide for configuring a new agents/private net. Some things I'd like to note on the way:
   - In scribe configs, start_block should be the deploy  block of the contract. This is especially important for executors since they need to rebuild the entire merkle tree from logs on each chain. In the past I've built a [`generate`](https://github.com/synapsecns/sanguine/blob/master/services/scribe/cmd/commands.go#L173) command that generates a scribe config based on a hardhat deploy folder using [this parser](https://pkg.go.dev/github.com/synapsecns/sanguine/ethergo/parser/hardhat). However this is not available for foundry. Do we know why this is/is there a way to fix this? (#1384)
[...]

PR #1383, opened 2023-09-15
remove unused scribe config params

trajan0x:
**Description**

Remove two global scribe params that appear to be unused found when working on #1381 

### Summary by CodeRabbit

---
- Refactor: Removed `RefreshRate` and `ConfirmationRefreshRate` fields from the `Config` struct in `services/scribe/config/config.go`. This change simplifies the configuration by eliminating unused refresh rates for updating block heights.
- Test: Updated tests in `services/scribe/config/config_test.go` and `services/scribe/service/scribe_test.go` to reflect the removal of the `RefreshRate` field. These changes ensure that our tests remain accurate and reliable after the refactor.
---

### DIFF
--- a/services/scribe/config/config.go
+++ b/services/scribe/config/config.go
@@ -14,12 +14,8 @@ import (
 type Config struct {
 	// Chains stores all chain information
 	Chains ChainConfigs `yaml:"chains"`
-	// RefreshRate is the rate at which the scribe will refresh the last block height in seconds.
-	RefreshRate uint `yaml:"refresh_rate"`
 	// RPCURL is the url of the omnirpc.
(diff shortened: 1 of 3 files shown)


trajan0x:
@nautsimon if neither of these two params are in use how often do we hit the latest block?

image

nautsimon:
> @nautsimon if neither of these two params are in use how often do we hit the latest block?
> 
> 
> 
> image
> 
> 

One of the changes to scribe was to set refresh rates at a contract level rather at a chain level. Refresh rate is an optional param that can be set for each contract, but will default to 1 second if not provided.

Once contracts are near head they are lumped into the big get logs query for multiple contracts and assume a 1 second refresh rate regardless of the set param.

PR #1387, opened 2023-09-18
Replace Create3 with ZeframLou/create3-factory

trajan0x:
**Description**

Addresses: https://github.com/synapsecns/sanguine/pull/1381#discussion_r1327651100 by using https://github.com/ZeframLou/create3-factory

### Summary by CodeRabbit

- Refactor: Updated the submodule URL for `packages/contracts-core/lib/create3-factory` to a new repository.
- New Feature: Added a new module `create3` in the `lib/create3-factory/src` directory.
- Refactor: Changed the contract name from `DeployCREATE3` to `DeployCREATE3Factory` and updated its import statements. The contract file is now part of a common deployer-utils package.
- Refactor: Updated the import path for the `CREATE3Factory` contract in `DeployerUtils.sol`, indicating a change in the contract's location.


aureliusbtc:
LGTM - but think we should keep a deploy script that imports the create3 lib in our scripts folder, so that it creates our standardized deployment JSON using DeployerUtils vs fully forcing the deployer to use zefram/create3 to deploy

trajan0x:
Can do this. One thing to consider here is @ChiTimesChi is working on a deployer-utils folder *soon* and I don't know how well tested the `Create3Factory` is (I've actually never ran it, I got it from #1312).

Happy to re-include for now

Issue #1397, opened 2023-09-22
`needs-go-generate` labels are not removed


ChiTimesChi:
An old PR could get a bunch of `needs-go-generate` labels if it is behind master, see for example:

https://github.com/synapsecns/sanguine/actions/runs/6265362203/job/17047824402

However, once the `master` branch is merged into the branch in question, the labels persist. This is due to the whole `Go Generate (Module Changes)` workflow being skipped on no changes:
https://github.com/synapsecns/sanguine/actions/runs/6265608504/job/17048082573

The desired behavior would be: when no changes are detected for a module, the `needs-go-generate` label for the module should be removed (if there is one).

trajan0x:
Fixed in #1399

PR #1382, opened 2023-09-15
SDK: cleanup

ChiTimesChi:
**Description**
Major refactor of Router SDK in preparation for RouterV2 and external Router Adapters.

High level overview of the changes:
- Introduced `Router` class that abstracts away the contract interactions and normalizes the data structures that different contracts are using, so that the rest of SDK (and its consumers) are dealing with unified types like `Query`.
  - `SynapseRouter` and `SynapseCCTPRouter` inherit from it to add support for corresponding contracts.
- Introduced `RouterSet` class that handles the interactions with Routers on a few chains (such as bridge quote).
  - `SynapseRouterSet` and `SynapseCCTPRouterSet` inherit from it to add support for corresponding contract sets.
- The SDK functions are split into a few types and implemented in `src/operations` to increase readability:
  - `src/operations/bridge` for bridging and bridge quoting,
  - `src/operations/pool` for pool inspection,
  - `src/operations/swap` for swapping and swap quoting.
- Cleaned up and adapted the test suite, added a bit more coverage.
  - Common test values now reside in `src/constants/testValues.ts`

The changes were done in a non-breaking way, so that the SDK consumers could update w/o adjusting their code.

**Additional context**
- This PR was moved from #1355 to include the @coderabbitai
- 663079e77be77ca9c1e15113a874adc2bf651ac2 was moved to this PR from #1356 so that a FE preview with the updated SDK could be generated within this PR **(needs to be reverted prior to merging)**

c552818cb344a049af8fc93620c4600c75042a77: [synapse-interface preview link ](https://sanguine-synapse-interface-6s4uc2wte-synapsecns.vercel.app)

cdc4c00b93c8b9af502651b743cf6422e5287827: [synapse-interface preview link ](https://sanguine-synapse-interface-4szhwyxjo-synapsecns.vercel.app)

### Summary by CodeRabbit

- New Feature: Added support for multiple chains with the introduction of `SynapseRouterSet` and `SynapseCCTPRouterSet` classes.
- New Feature: Introduced new operations for bridging, swapping, and pool interactions in the SynapseSDK module.
- New Feature: Added `SynapseRouter` and `SynapseCCTPRouter` classes to interact with respective contracts on different chains.
- Refactor: Reorganized constants and test values into separate modules for better code organization.
- Test: Added comprehensive test suites for new classes and functions to ensure correct functionality.
- Chore: Updated jest configuration to increase test timeout.

[...]

Issue #579, opened 2023-02-21
Needs go generate tag inaccurate 


CryptoMaxPlanck:
In PRs such #576 , PR's that might have some go changes, and revert them do not remove the go generate tag.

PR #1399, opened 2023-09-25
Fix git-changes-action when module was unchanged

trajan0x:
**Description**

As #1397 & #579 pointed out if a module is changed and then unchanged, the needs-go-generate action is not removed (this is blocking #1382 rn). For this reason, this pr:

- consolidates actions for `include_deps`=true and `include_deps`=false. This will also help with #1390 by halving api requests by git-changes-action. Further work could be done here by consolidating goreleaser into `go.yml` the future 
- Remove unchanged module labels. This is honestly kind of a pain and could lead to api rate limiting issues. The way we're going to deal with this is by starting an action for each label, but only running it if the event is a pull request and [this condition](https://stackoverflow.com/a/59588725) is met. Once this is done we can also add this to the needs-go-generate check to reduce the number of api calls to remove a lebl further helping w/ #1390 

Lesson from #1400: paralell removal is not possible

## Future Work

 -  This will still break is a module is deleted entirely and the `needs-go-generate-x...` label is on a pr , but I believe this is an acceptable edge case. If we want we can rule this out by adding the number of unchanged modules to changed modules and dividing by total and removing excess, but this seems gratuitous at present

 - One other thing I figured out is that for some reason these generation actions are ran on a `push` rather than a `pull_request` level creating the need for two additional api requests. We should figure out if there was a reason for this and fix it if not
 - This spins up *a lot* of jobs, but their total duration is on average ~3 seconds, and they usually don't do anything.
 - Run in a single job rather than one per unchanged. This is mostly aesthetic at this time since these hardly have any overhead. see #1400 for why this is neccesary, removing labels in paralell doesn't work)
 - Labels can't be removed in paralell rn, so the other jobs are broken

### Summary by CodeRabbit

- Chore: Updated various GitHub Actions workflows, improving their efficiency and maintainability. This includes version updates for actions such as `actions/stale`, `actions/checkout`, `actions/setup-go`, and `actions/setup-python`.
- Refactor: Enhanced the `git-changes-action` by introducing a new function `outputModuleChanges` to improve modularity and output structure for GitHub Actions.
- Bug Fix: Removed unused steps in the `fail-on-needs-generate.yaml` workflow, enhancing its performance.
- New Feature: Added a job to remove the "needs-go-generate" label from unused jobs in the `go.yml` workflow.
- Documentation: Updated the README for `git-changes-action` to reflect recent changes and improvements.

### DIFF
--- a/.github/workflows/close-stale.yml
+++ b/.github/workflows/close-stale.yml
@@ -2,12 +2,16 @@ name: 'Close stale issues and PRs'
 on:
   schedule:
     - cron: '30 1 * * *'
+  workflow_dispatch:
+  push:
+    paths:
+      - '.github/workflows/close-stale.yml'
 
 jobs:
   stale:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/stale@v4
+      - uses: actions/stale@v8
         with:
           stale-pr-message: 'This PR is stale because it has been open 14 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
           exempt-pr-labels: exempt-stale
--- a/.github/workflows/fail-on-needs-generate.yaml
+++ b/.github/workflows/fail-on-needs-generate.yaml
@@ -10,7 +10,6 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v4
       - name: Fail on needs-go-generate
         run: echo Failed, needs go generate. && exit 1
   check-generate-yarn:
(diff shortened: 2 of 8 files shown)


trajan0x:
issue as of ebe5e58092a9a805dc33b6272426119e30a5baa6 (for my ref)
image

trajan0x:
testing this out again following this event: https://github.com/synapsecns/sanguine/pull/1399#event-10465688166

PR #1436, opened 2023-10-06
Fix: ToB-14 (tips with eoa recipient)

ChiTimesChi:
**Description**
- Fixed ToB-14 issue by ensuring that `executeBaseMessage` doesn't revert if the recipient is EOA.
- This is achieved by using a new library for performing such "safe calls" instead of `try/catch` block that ends up in revert, if the target contract doesn't contain any code.

### Summary by CodeRabbit

- New Feature: Introduced `SafeCall` library in `contracts-core` package to safely perform calls to recipient contracts, improving the robustness of contract interactions.
- Test: Added comprehensive test coverage for the `SafeCall` library in `SafeCallTest`, ensuring its correct functionality across various scenarios including out-of-gas and revert situations.
- Test: Enhanced `ExecutionHubTest` with a new function `test_execute_base_recipientEOA` to validate execution of base messages with externally owned account recipients.
- Chore: Created mock contracts `CalleeMock` and `CalleeReturnDataMock` for testing purposes.

### DIFF
--- a/packages/contracts-core/contracts/libs/SafeCall.sol
+++ b/packages/contracts-core/contracts/libs/SafeCall.sol
@@ -1,0 +1,30 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+/// @notice Library for performing safe calls to the recipients.
+/// Inspired by https://github.com/nomad-xyz/ExcessivelySafeCall
+library SafeCall {
+    /// @notice Performs a call to the recipient using the provided gas limit, msg.value and payload
+    /// in a safe way:
+    /// - If the recipient is not a contract, false is returned instead of reverting.
+    /// - If the recipient call reverts for any reason, false is returned instead of reverting.
+    /// - If the recipient call succeeds, true is returned, and any returned data is ignored.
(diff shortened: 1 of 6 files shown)


packages/contracts-core/contracts/libs/SafeCall.sol
```diff
@@ -0,0 +1,30 @@
...
+        internal
+        returns (bool success)
+    {
+        // Exit early if the recipient is not a contract
+        if (recipient.code.length == 0) return false;
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            // Perform the call to the recipient, while ignoring any returned data
+            // call(g, a, v, in, insize, out, outsize) => returns 0 on error, 1 on success
+            success := call(gasLimit, recipient, msgValue, add(payload, 0x20), mload(payload), 0, 0)
```
mikeyrf, line 27: Any reason for assembly vs an [OZ-like approach](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Address.sol#L87)?

mikeyrf (approved):
lgtm

Issue #1429, opened 2023-10-06
Research the possibility of overflows


ChiTimesChi:
We are using the tight packing wherever possible in the contracts, as this both reduces the storage costs, and trims the sizes of the merkle trees.

Need to do the research whether the following constraints are adequate:
1. Attestation
  - `uint32 nonce`
  - `uint40 blockNumber`
  - `uint40 timestamp`
2. State
  - `uint32 origin`
  - `uint32 nonce`
  - `uint40 blockNumber`
  - `uint40 timestamp`
3. Header
  - `uint32 origin, uitn32 destination`
  - `uint32 nonce`
  - `uint32 optimisticPeriod`
4. Request
  - `uint96 gasDrop`
  - `uint64 gasLimit`
5. Tips
  - `uint64 summitTip, attestationTip, executionTip, deliveryTip`
6. Merkle Tree max height
  - **Origin Merkle Tree:** 32
  - **Agent Merkle Tree:** 32
  - **Snapshot Merkle Tree:** 6

ChiTimesChi:
We are also using a few `uint32` to store a few different "index" values:
- Index of agent in the Agent Merkle Tree: fine as long as max height is 32.
- Submitted snapshot roots in `Destination`: fine as long as attestation nonces fit into `uint32` (each attestation could only be saved once in `Destination`)
- Index of origin states in `Origin`: fine as long as origin nonces fit into `uint32`.

PR #1445, opened 2023-10-11
Fix: ToB-05 (destination chain agent root)

ChiTimesChi:
**Description**
- Fixed ToB-05 by allowing the contract owner to set the Agent Merkle Root in `LightManager`. This could be used to resolve disputes or update the agent root in general, if no active Notaries exist under the current one.
  - Function uses the same "no fresh data for X minutes" check as `resolveStuckDispute` used to have.
- `resolveStuckDispute` is moved to `BondingManager`, as it is no longer necessary in `LightManager`.
- `resolveStuckDispute` was renamed into `resolveDisputeWhenStuck` for consistency.

### Summary by CodeRabbit

- New Feature: Introduced a new mechanism to handle stuck disputes. Contract owners can now manually resolve these disputes under certain conditions, improving the reliability of the system.
- New Feature: Added the ability for contract owners to propose, cancel, and resolve a new agent root when the contract is stuck, enhancing control and flexibility.
- New Feature: Added new events related to agent root management, increasing transparency and visibility into the agent root resolution process.
- Test: Added comprehensive test cases for the new features, ensuring robustness and reliability.
- Refactor: Streamlined error handling with new error types and removed redundant ones, improving code readability and maintainability.
- Chore: Updated import statements and constants to align with the new changes.


packages/contracts-core/contracts/manager/AgentManager.sol
```diff
@@ -65,6 +64,13 @@
         _;
     }
 
+    modifier onlyWhenStuck() {
+        // Check if there has been no fresh data from the Notaries for a while.
+        (uint40 snapRootTime,,) = InterfaceDestination(destination).destStatus();
+        if (block.timestamp < FRESH_DATA_TIMEOUT + snapRootTime) revert NotStuck();
+        _;
+    }
```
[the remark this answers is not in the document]
trajan0x: Wonder if we wanna open an issue on slither about this?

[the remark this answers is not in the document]
ChiTimesChi: Yeah, I will work on the minimal reproduction later this week.

packages/contracts-core/contracts/manager/LightManager.sol
```diff
@@ -53,6 +55,13 @@ contract LightManager is AgentManager, InterfaceLightManager {
         __Ownable_init();
     }
 
+    // ════════════════════════════════════════════════ OWNER ONLY ═════════════════════════════════════════════════════
+
+    /// @inheritdoc InterfaceLightManager
+    function setAgentRootWhenStuck(bytes32 agentRoot_) external onlyOwner onlyWhenStuck {
```
mikeyrf, line 61: ToB-05 suggests a two step function with a time delay, so consider adding a second step similar to OZ's [`OwnableTwoStep`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable2Step.sol).

mikeyrf (approved):
One suggestion potentially omitted, but otherwise lgtm. Up to you if you want to edit for ToB two step functionality.

packages/contracts-core/contracts/manager/AgentManager.sol
```diff
@@ -65,6 +64,13 @@ abstract contract AgentManager is MessagingBase, AgentManagerEvents, IAgentManag
         _;
     }
 
+    modifier onlyWhenStuck() {
+        // Check if there has been no fresh data from the Notaries for a while.
+        (uint40 snapRootTime,,) = InterfaceDestination(destination).destStatus();
+        if (block.timestamp < FRESH_DATA_TIMEOUT + snapRootTime) revert NotStuck();
```
mikeyrf, line 70: `snapRootTime` from destination is cast to a `uint40`. if `block.timestamp > type(uint40).max`, the only when stuck check won't be valid (always returns `true` as stuck).

should cast `block.timestamp` here to `uint40(block.timestamp)` to be consistent and avoid this edge case.

mikeyrf (changes requested):

packages/contracts-core/contracts/manager/AgentManager.sol
ChiTimesChi: I would argue that it's the responsibility of the contract that is producing data (Destination) rather than the one that is consuming it (AgentManager). 

The #1457 fixes the safe casts so that it's not possible to save the corrupted data, and the amount of bits to allocate to timestamps is tracked in #1429

mikeyrf (approved):

packages/contracts-core/contracts/manager/LightManager.sol
ChiTimesChi: Could go for two-step process here, agreed

ChiTimesChi:
@mikeyrf the two-step process is now implemented. Will fix the merge conflict once I merge some of the other PRs into master.

packages/contracts-core/contracts/manager/LightManager.sol
```diff
@@ -53,6 +68,30 @@ contract LightManager is AgentManager, InterfaceLightManager {
         __Ownable_init();
     }
 
+    // ════════════════════════════════════════════════ OWNER ONLY ═════════════════════════════════════════════════════
+
+    /// @inheritdoc InterfaceLightManager
+    function proposeAgentRootWhenStuck(bytes32 agentRoot_) external onlyOwner onlyWhenStuck {
+        // Update the proposed agent root, clear the timer if the root is empty
+        _proposedAgentRoot = agentRoot_;
+        _agentRootProposedAt = agentRoot_ == 0 ? 0 : block.timestamp;
```
mikeyrf, line 77: what's the logic for allowing owner to propose `agentRoot_ = 0` here if would then revert below when attempt to call `resolveProposedAgentRoot`?

mikeyrf (approved):
lgtm. left one extra question for you in case.

packages/contracts-core/contracts/manager/LightManager.sol
ChiTimesChi: That's a good point. It was done for the purpose of cancelling the previously proposed one, but it's much cleaner to have a separate function for that. Plus might need to be able to cancel the proposed root if the chain became unstuck naturally.

Will work on a quick fix for that.

mikeyrf (approved):

aureliusbtc (approved):

PR #1449, opened 2023-10-12
Fix: ToB-08 (indefinite stuck disputes)

ChiTimesChi:
**Description**
- Fixed ToB-08 by forcing the every Notary to submit the attestations in the ascending nonce order. This means that they are forced to submit a fresher attestation each time, which will contain the fresher agent root, inevitably resulting in the dispute resolution.
- The order enforcing is implemented on the individual level rather than overall in order to prevent malicious Notaries from submitting attestation with nonce that doesn't exist and effectively disabling other Notaries. Also prevents DoS attack by frontrunning attestation submission with useless attestation (having a single chain state for instance) having a higher nonce.

### Summary by CodeRabbit

**New Feature:**
- Added nonce tracking for attestations in `Destination.sol`. This feature ensures that each attestation has a unique nonce, preventing replay attacks and ensuring the integrity of the attestation process.

**Test:**
- Added multiple test cases in `Destination.t.sol` to verify the correct handling of attestation nonces. These tests cover scenarios such as lower nonce, same nonce, and different notary with lower or same nonce.

### DIFF
--- a/packages/contracts-core/contracts/Destination.sol
+++ b/packages/contracts-core/contracts/Destination.sol
@@ -52,6 +52,9 @@ contract Destination is ExecutionHub, DestinationEvents, InterfaceDestination {
     /// @inheritdoc InterfaceDestination
     DestinationStatus public destStatus;
 
+    /// @inheritdoc InterfaceDestination
+    mapping(uint32 => uint32) public lastAttestationNonce;
+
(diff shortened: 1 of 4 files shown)


aureliusbtc (approved):

trajan0x (approved):
clever

PR #1475, opened 2023-10-20
Add block height endpoint to explorer api

nautsimon:
**Description**
Provides a new api endpoint that returns the current block height status for contracts in the explorer data base.

Takes an array of objects of the following type
```
type ContractQuery {
  chainID: Int
  type: ContractType (Enum of string BRIDGE or CCTP)
}
```
 
 and returns an array of objects in of the following type in ascending block height order
 
 ```
 type BlockHeight {
  chainID: Int
  type: ContractType (Enum of string BRIDGE or CCTP)
  blockNumber: Int
}
```

### DIFF
--- a/services/explorer/graphql/server/graph/queries.resolvers.go
+++ b/services/explorer/graphql/server/graph/queries.resolvers.go
@@ -427,6 +427,35 @@ func (r *queryResolver) GetDestinationBridgeTx(ctx context.Context, chainID int,
 	return results, nil
 }
 
+// GetBlockHeight is the resolver for the getBlockHeight field.
+func (r *queryResolver) GetBlockHeight(ctx context.Context, contracts []*model.ContractQuery) ([]*model.BlockHeight, error) {
+	// Generate string for right side of IN clause
+	var contractString string
+	contractTypeMap := make(map[string]*model.ContractType)
+	for i, contract := range contracts {
+		if contract.ChainID == nil || contract.Type == nil {
+			return nil, fmt.Errorf("chainID and type must be set")
+		}
+		contractAddr, err := r.getContractAddressFromType(uint32(*contract.ChainID), *contract.Type)
+		if err != nil {
+			return nil, fmt.Errorf("could not get contract address from type %s, %w", contract.Type, err)
+		}
+		contractTypeMap[contractAddr] = contract.Type
+		if i == 0 {
+			contractString += fmt.Sprintf("('%s', %d)", contractAddr, *contract.ChainID)
+		} else {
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: services/explorer/graphql/server/graph/queryutils.go [hunk 1]
```diff
@@ -1835,3 +1835,17 @@ func (r *queryResolver) checkIfChainIDExists(chainIDNeeded uint32, bridgeType mo
...
+	}
+	switch contractType {
+	case model.ContractTypeBridge:
+		return r.Config.Chains[chainID].Contracts.Bridge, nil
+	case model.ContractTypeCctp:
+		return r.Config.Chains[chainID].Contracts.CCTP, nil
+	default:
+		return "", fmt.Errorf("contract type not supported")
+	}
+}
```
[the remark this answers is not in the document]
trajan0x: @nautsimon was going to comment same. I know it's a bit annoying but

```go
chain, ok := r.Config.Chains[chainID].
if !ok {
  // do something
}
if chain.Contracts == nil (idk if this is pointer or not, but you get the idea), if it's not pointer you can ignore but we should make sure that a nil panic here can't crash the server
```

services/explorer/graphql/server/graph/schema/types.graphql
```diff
@@ -225,3 +224,21 @@ enum KappaStatus{
   PENDING
   UNKNOWN
 }
+
+input ContractQuery {
```
trajan0x, line 228: if one of these is required (I think both may be?) should add a `!` (note: coordinate w/ @bigboydiamonds I thikn this may require a fe change)

services/explorer/graphql/server/graph/queries.resolvers.go
```diff
@@ -427,6 +427,35 @@ func (r *queryResolver) GetDestinationBridgeTx(ctx context.Context, chainID int,
 	return results, nil
 }
 
+// GetBlockHeight is the resolver for the getBlockHeight field.
+func (r *queryResolver) GetBlockHeight(ctx context.Context, contracts []*model.ContractQuery) ([]*model.BlockHeight, error) {
+	// Generate string for right side of IN clause
+	var contractString string
+	contractTypeMap := make(map[string]*model.ContractType)
+	for i, contract := range contracts {
+		if contract.ChainID == nil || contract.Type == nil {
```
trajan0x, line 436: is there any scenario where contrac tis also nil here?

trajan0x (changes requested):

services/explorer/graphql/server/graph/schema/types.graphql
trajan0x: #1482

PR #1476, opened 2023-10-20
Make agents tests faster

dwasse:
**Description**
Set anvil block time to 1 second which should make event processing more robust (instead of having to rely on bumping blocks with dummy events).

**Additional context**
Decrease amount of retries in CI to notify faster if an agents test has actually broken.

### Summary by CodeRabbit

- Test Improvement: Reduced the maximum number of retries for test execution from 5 to 3, enhancing the efficiency of our testing process.
- Consistency Enhancement: Ensured uniform block time across different instances in our testing suite, leading to more reliable and consistent test results.

These changes primarily improve our internal testing procedures, making them more efficient and reliable. While these updates may not directly impact end-users, they help us deliver higher quality software more quickly.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -183,7 +183,7 @@ jobs:
       - name: Test
         run: |
           set -e
-          max_num_retries=5
+          max_num_retries=3
           echo "mode: atomic" > coverage.txt
 
           # Loop through each package and run tests, with package level retries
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -192,7 +192,7 @@ jobs:
             retries=0
 
             while [ $retries -lt $max_num_retries ]; do
-              if $PYROSCOPE_PATH exec --apiKey=${{ secrets.PYROSCOPE_CLOUD_TOKEN }} -- gotestsum --rerun-fails=5 --packages="$pkg" --format standard-verbose -- -v -coverpkg="$pkg" -coverprofile=profile.cov $pkg; then
+              if $PYROSCOPE_PATH exec --apiKey=${{ secrets.PYROSCOPE_CLOUD_TOKEN }} -- gotestsum --rerun-fails=3 --packages="$pkg" --format standard-verbose -- -v -coverpkg="$pkg" -coverprofile=profile.cov $pkg; then
```
trajan0x, line 195: @dwasse are we sure this isn't 3*3=9 retries? Seems like 3 per package a total of 3 times: https://github.com/gotestyourself/gotestsum#re-running-failed-tests

If that's intended behavior/we don't know how to avoid we should document this

trajan0x (changes requested):

.github/workflows/go.yml
dwasse: Yeah the intention is allow retrying all agents tests as well as individual packages. Perhaps we can lower even more? Really just trying to find the appropriate setting that will allow broken tests to fail without having to wait a long time.

PR #1684, opened 2023-12-20
Add `FeePricer` to quoter

dwasse:
Adds `FeePricer` for setting the `FixedFee` value in the quoter.


trajan0x:
Managed to identify the issue behind the failing e2e tests. `feePricer` is failing to recognize the token


image

services/rfq/e2e/setup_test.go
```diff
@@ -194,10 +196,24 @@ func (i *IntegrationSuite) setupRelayer() {
 			originBackendChainID: {
 				Bridge:        i.manager.Get(i.GetTestContext(), i.originBackend, testutil.FastBridgeType).Address().String(),
 				Confirmations: 0,
+				Tokens: map[string]relconfig.TokenConfig{
+					"ETH": relconfig.TokenConfig{
```
trajan0x, line 200: Not sure we actually need gas token symbol here, but ok.

Think the reality is, before e deploy (or slightly after) config is going to need a complete overhaul in #1650, just don't think this pr is the appropriate place to do that.

My recommendations or how to go about this is:

1. Convert all current references to pieces of the `Config` struct to Getters similiar to [Submitter IConfig](https://pkg.go.dev/github.com/synapsecns/sanguine/ethergo@v0.1.1/submitter/config#IConfig)
2. Reference those
3. Rewrite the configs
4. Fix the e2e tests

trajan0x (approved):
LGTM

PR #1687, opened 2023-12-21
[ignore] Debug/new rfq fixed fee log

trajan0x:
Just using for diff tracking:

https://github.com/synapsecns/sanguine/pull/1684#issuecomment-1865529464


services/rfq/relayer/quoter/quoter.go
```diff
@@ -164,6 +170,23 @@ func (m *Manager) GenerateQuotes(ctx context.Context, chainID int, address commo
 				}
 				destToken, err := m.feePricer.GetTokenName(uint32(chainID), address.Hex())
 				if err != nil {
+
```
trajan0x, line 173: everything but ctx fix above tertiary

PR #1698, opened 2023-12-21
Fix Call Context Endpoint

trajan0x:
**Description**

Fix the c`CallContext` endpoint


trajan0x:
Will fix conflict

aureliusbtc:
conflict needs to be fixed

PR #1726, opened 2024-01-02
Add `DestTxHash` field and use it in relayer `Prove()` call

dwasse:
**Description**

Add the `DestTxHash` field to `QuoteRequest` and `RequestForQuote` models, and set this field when `RelayComplete` status is set.


trajan0x:
Gonna get this in prior to #1725

trajan0x:
LGTM. Might consider a test here, depending on how comfy you feel

PR #1858, opened 2024-01-13
fe/refactor-activity-remove-pending

bigboydiamonds:
## Summary by CodeRabbit

- **Refactor**
  - Optimized transaction management to focus on historical data.
  - Removed outdated components and hooks for cleaner architecture.

06edcc499819cc513ecc2a669c6e77477f92129b: [synapse-interface preview link ](https://sanguine-synapse-interface-afh85h4f3-synapsecns.vercel.app)
ee625288ea8e4c7a99e04fb1d5f197e75597d011: [synapse-interface preview link ](https://sanguine-synapse-interface-6a6jhdv50-synapsecns.vercel.app)
4bfa61dab1ca6a8ab9a9450f7559faeb9c683e84: [synapse-interface preview link ](https://sanguine-synapse-interface-8hqil056g-synapsecns.vercel.app)
0f4237726e19e258179221fb264af594615a575f: [synapse-interface preview link ](https://sanguine-synapse-interface-2hpjsaxcz-synapsecns.vercel.app)
907804c32a6e49035e75fcd6b50f377892029e88: [synapse-interface preview link ](https://sanguine-synapse-interface-h0w8eioqp-synapsecns.vercel.app)


packages/synapse-interface/store/reducer.ts
```diff
@@ -27,7 +27,7 @@ const persistedReducers = {
 export const storageKey: string = 'synapse-interface'
 
 export const persistConfig: PersistConfig<AppState> = {
-  version: 1, // upgrade to reset cache when updated data structures throw errors
+  version: 2, // upgrade to reset cache when updated data structures throw errors
```
abtestingalpha, line 30: Do we need to update this here?

abtestingalpha (changes requested):

packages/synapse-interface/store/reducer.ts
bigboydiamonds: Maybe not, added just in case since we're removing state from a persisted (transaction) slice. Can remove if you think it's unnecessary

bigboydiamonds: Checked this out and it looks like we don't have cache collision issues. Given we've removed all instances of the state both from reducers, functions, and around the app, there shouldn't be an opportunity for a crash.

Even if we don't upgrade persist version, users with an previous version of state won't have any code that would trigger a white screen issue.

The primary risk here is if we eventually end up adding back a transaction state value that shared the same names as the ones I just removed in this PR. To be on the safe side, we can still continue with upgrading persist version so we completely cache bust on all users. Only thing here is that we're going to wipe User's existing tx's when that happens

PR #1867, opened 2024-01-15
SDK: rework docs

ChiTimesChi:
**Description**
Updates SDK docs.


Defi-Moses (approved):
Docs look good to me. Added a small tag for slippage and deadline to clarify that these parameters aren't necessary (but slippage is recommended). 

In the future we can explore making the graphics easier to read/ cleaner but this is more than enough of an improvement to help developers.

PR #1871, opened 2024-01-16
Sdk: include rfq protocol fees

ChiTimesChi:
**Description**
This PR adds support for protocol fees in the `FastBridge` contract. 

## Summary by CodeRabbit

- **New Features**
	- Introduced a feature to retrieve the protocol fee rate within the routing system.
	- Updated routing calculations to include protocol fee adjustments, ensuring accurate cost representation during transactions.

### DIFF
--- a/packages/sdk-router/src/rfq/fastBridgeRouterSet.ts
+++ b/packages/sdk-router/src/rfq/fastBridgeRouterSet.ts
@@ -112,13 +112,19 @@ export class FastBridgeRouterSet extends SynapseModuleSet {
       amountIn,
       allQuotes
     )
+    const protocolFeeRate = await this.getFastBridgeRouter(
+      originChainId
+    ).getProtocolFeeRate()
     return filteredQuotes
       .map(({ quote, originQuery }) => ({
         quote,
         originQuery,
-        // Apply quote to the proceeds of the origin swap
+        // Apply quote to the proceeds of the origin swap with protocol fee applied
         // TODO: handle optional gas airdrop pricing
-        destAmountOut: applyQuote(quote, originQuery.minAmountOut),
+        destAmountOut: applyQuote(
+          quote,
+          this.applyProtocolFeeRate(originQuery.minAmountOut, protocolFeeRate)
+        ),
       }))
       .filter(({ destAmountOut }) => destAmountOut.gt(0))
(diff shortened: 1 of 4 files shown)


packages/sdk-router/src/rfq/fastBridgeRouterSet.ts
```diff
@@ -112,13 +112,19 @@ export class FastBridgeRouterSet extends SynapseModuleSet {
...
       .map(({ quote, originQuery }) => ({
         quote,
         originQuery,
-        // Apply quote to the proceeds of the origin swap
+        // Apply quote to the proceeds of the origin swap with protocol fee applied
         // TODO: handle optional gas airdrop pricing
-        destAmountOut: applyQuote(quote, originQuery.minAmountOut),
+        destAmountOut: applyQuote(
+          quote,
+          this.applyProtocolFeeRate(originQuery.minAmountOut, protocolFeeRate)
```
mikeyrf, line 126: Applied on origin amount tho so is it relevant for FE as only used by caller of `claim()` which will be relayer?

mikeyrf (approved):
One question but otherwise lgtm

packages/sdk-router/src/rfq/fastBridgeRouterSet.ts
mikeyrf: Other consideration would be on `refund()`, bridge sends back to origin sender `transaction.originAmount + transaction.originFeeAmount` so protocol doesn't take a fee on unsuccessful bridges, if that affects SDK in any way.

ChiTimesChi: I am replicating the internal logic of `FastBrdige.bridge()` wrt the origin amount used for the cross-chain swap:
https://github.com/synapsecns/sanguine/blob/1b95502e029e6e14b588b5dff8a9c3ad0d4c76ec/packages/contracts-rfq/contracts/FastBridge.sol#L99

This should be the amount that the off-chain Relayer will recognize, and therefore to get an accurate final quote for the user, the Relayer's quote is applied to the amount after the protocol fees.

Does this make sense?

mikeyrf: Yup! That makes sense then

PR #1946, opened 2024-01-29
Add ARB cap per address in STIP relayer

dwasse:
**Description**
Adds a cap of ARB per address in the STIP relayer. This is done by querying all STIP transactions for a given address and summing the `arb_amount_rebated` field.

Note that the default cap is set to 2000 ARB, but config value will override.

### DIFF
--- a/services/stiprelayer/db/sql/base/store.go
+++ b/services/stiprelayer/db/sql/base/store.go
@@ -42,7 +42,7 @@ func (s *Store) GetTotalArbRebated(ctx context.Context, address string) (*big.In
 		if !ok {
 			return nil, fmt.Errorf("failed to convert arb amount rebated to number")
 		}
-		totalRebated = totalRebated.Add(totalRebated, rebatedAmount)
+		totalRebated.Add(totalRebated, rebatedAmount)
 	}
 	return totalRebated, nil
 }
--- a/services/stiprelayer/db/sql/base/store.go
+++ b/services/stiprelayer/db/sql/base/store.go
@@ -2,6 +2,8 @@ package base
 
 import (
 	"context"
+	"fmt"
+	"math/big"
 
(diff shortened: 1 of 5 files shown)


services/stiprelayer/db/sql/base/store.go
```diff
@@ -20,6 +22,31 @@ func (s *Store) GetSTIPTransactionsNotRebated(ctx context.Context) ([]*db.STIPTr
...
+		Where("address = ?", address).
+		Find(&stipTransactions)
+	if result.Error != nil {
+		return nil, result.Error
+	}
+
+	// Compute the sum of arb rebated across all transactions
+	totalRebated := big.NewInt(0)
+	for _, stipTransaction := range stipTransactions {
+		rebatedAmount, ok := new(big.Int).SetString(stipTransaction.ArbAmountRebated, 10)
```
aureliusbtc, line 41: I don't think this converts from 18 decimals accurately, maybe:

```
weiInOneEther := big.NewInt(0).Exp(big.NewInt(10), big.NewInt(18), nil)
rebatedAmountWei, _ := new(big.Int).SetString(stipTransaction.ArbAmountRebated, 10)
rebatedAmountEther := new(big.Float).Quo(new(big.Float).SetInt(rebatedAmountWei), new(big.Float).SetInt(weiInOneEther))
```

dwasse: I thought the `arb_amount_rebated` field was already in wei units, hence the `String` type? Since we are dealing with wei units in relayer I dont think this conversion is necessary?

aureliusbtc: yep you're right lgtm

aureliusbtc (approved):

PR #2035, opened 2024-02-14
Fe/transaction balance fetch

bigboydiamonds:
**Description**

**Additional context**

**Metadata**
- Fixes #[Link to Issue]

### DIFF
--- a/packages/synapse-interface/pages/state-managed-bridge/index.tsx
+++ b/packages/synapse-interface/pages/state-managed-bridge/index.tsx
@@ -73,6 +73,7 @@ import {
   fetchEthPrice,
   fetchGmxPrice,
 } from '@/slices/priceDataSlice'
+import { isTransactionReceiptError } from '@/utils/isTransactionReceiptError'
 
 const StateManagedBridge = () => {
   const { address } = useAccount()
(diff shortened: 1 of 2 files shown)


packages/synapse-interface/pages/state-managed-bridge/index.tsx
```diff
@@ -473,6 +474,21 @@ const StateManagedBridge = () => {
       dispatch(removePendingBridgeTransaction(currentTimestamp))
       console.log('Error executing bridge', error)
       toast.dismiss(pendingPopup)
+
+      /** Fetch balances if await transaction receipt times out */
+      if (isTransactionReceiptError(error)) {
+        console.log(
+          'isTransactionReceiptError: ',
+          isTransactionReceiptError(error)
```
abtestingalpha, line 482: This always returns true if it's in this `if` block here, correct? Can just log the string without calling the `isTransactionReceiptError(error)` fn again

bigboydiamonds: yeah that's right - only meant to include log for testing purposes though, was planning on removing after reviews - doing so now

PR #2262, opened 2024-03-13
do not rebalance when rebalance amount is 0

trajan0x:
**Description**

Fixes rebalancing

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved the rebalance process by adding a check to prevent unnecessary operations when the rebalance amount is zero.


trajan0x:
Closing in favor of #2263

PR #2301, opened 2024-03-15
Feat/interchain batching prototype backend

trajan0x:
**Description**

Add backend implementation for interchain batching

Side note: for operational reasons I'd recommend we follow one of the two courses wrt merging (post #2303):

1. Merge (*don't squash and merge* #2239 into #2234), then Merge this into #2234 after changing the target
2. Squash and Merge into #2239 then merge or squash and merge that into #2234


committee/db/base/signrequest.go
```diff
@@ -18,8 +18,9 @@ type VerificationRequest struct {
 	TXHash string `gorm:"column:tx_hash;index;size:256"`
 	// TransactionID is the id of the transaction
 	TransactionID string `gorm:"column:transaction_id;index;size:256;primaryKey"`
-	// DataHash is the hash of the data
-	Entry string `gorm:"column:entry"`
+	// Entry is the entry to be signed
+	// TODO: consider renaming to batch
```
ChiTimesChi, line 22: Let's rename this? Feels like the DB will be dropped anyway, so it shouldn't be a blocker.

PR #2432, opened 2024-04-02
Submitter/dynamic tx fees [goreleaser]

aureliusbtc:
**Description**

Misc fixes for #2410

## Summary by CodeRabbit

- **New Features**
    - Updated gas price management for transactions with dynamic fee adjustments.
    - Introduced an environment variable `METRICS_ALWAYS_SYNC` to control metric export behavior.
- **Bug Fixes**
    - Corrected transaction processing logic to accurately mark transactions before a specific nonce as replaced or confirmed.
    - Improved market-based gas price calculations for better EIP-1559 support.
- **Documentation**
    - Updated README to include details on the `METRICS_ALWAYS_SYNC` environment variable.
- **Refactor**
    - Enhanced efficiency by renaming and optimizing `MarkAllBeforeOrAtNonceReplacedOrConfirmed` to `MarkAllBeforeNonceReplacedOrConfirmed`.
- **Style**
    - Ensured consistency in error logging messages across various services.

PR #2462, opened 2024-04-05
Submitter: setGasPrice() refactor

dwasse:
## Summary by CodeRabbit

- **New Features**
	- Improved handling of gas price settings to support dynamic pricing, including EIP-1559 transactions.
	- Added functionality to adjust gas prices based on market conditions and transaction types.

- **Bug Fixes**
	- Fixed a potential issue where `nil` gas prices could cause errors by adding a `nil` check.

- **Refactor**
	- Renamed `BaseGasPrice` to `MinGasPrice` across several files for clarity.
	- Updated method names and comments in the `IConfig` interface to reflect the new gas price terminology.

- **Tests**
	- Enhanced testing for dynamic and legacy transaction gas pricing scenarios.

### DIFF
--- a/ethergo/submitter/submitter_test.go
+++ b/ethergo/submitter/submitter_test.go
@@ -41,7 +41,7 @@ func (s *SubmitterSuite) TestSetGasPrice() {
 	s.Require().NoError(err)
 
 	maxGasPrice := big.NewInt(1000 * params.GWei)
-	minGasPrice := big.NewInt(10 * params.GWei)
+	minGasPrice := big.NewInt(1 * params.GWei)
 	cfg := &config.Config{
 		Chains: map[int]config.ChainConfig{
 			int(legacyChainID.Int64()): {
@@ -114,13 +114,13 @@ func (s *SubmitterSuite) TestSetGasPrice() {
 
 	s.Run("DynamicTx:BelowMin", func() {
 		resetTransactors()
-		gasPrice := big.NewInt(2 * params.GWei)
-		gasTipCap := big.NewInt(1 * params.GWei)
+		gasPrice := big.NewInt(0.5 * params.GWei)
+		gasTipCap := big.NewInt(0)
 		client.On(testsuite.GetFunctionName(client.SuggestGasPrice), mock.Anything).Once().Return(gasPrice, nil)
 		client.On(testsuite.GetFunctionName(client.SuggestGasTipCap), mock.Anything).Once().Return(gasTipCap, nil)
 		err = ts.SetGasPrice(s.GetTestContext(), client, dynamicTransactor, dynamicChainID, nil)
 		s.Require().NoError(err)
-		assertGasValues(dynamicTransactor, nil, minGasPrice, minGasPrice)
+		assertGasValues(dynamicTransactor, nil, minGasPrice, big.NewInt(10*params.Wei))
 	})
 
 	s.Run("LegacyTx:AboveMax", func() {


ethergo/submitter/config/config.go
```diff
@@ -67,8 +67,8 @@ const (
 // DefaultMaxPrice is the default max price of a tx.
 var DefaultMaxPrice = big.NewInt(500 * params.GWei)
 
-// DefaultBaseGasPrice is the default max price of a tx.
-var DefaultBaseGasPrice = big.NewInt(1 * params.GWei)
+// DefaultMinGasPrice is the default min price of a tx.
+var DefaultMinGasPrice = big.NewInt(10 * params.GWei)
```
ChiTimesChi, line 71: Min gas price of 10 Gwei is an overshoot.

ethergo/submitter/submitter.go
```diff
@@ -346,119 +346,211 @@ func (t *txSubmitterImpl) txTypeForChain(chainID *big.Int) (txType uint8) {
...
-			transactor.GasFeeCap = maxPrice
-			shouldBump = false
-			span.AddEvent("not bumping fee cap since max price is reached")
+	} else {
+		if useDynamic {
+			if transactor.GasFeeCap == nil || transactor.GasFeeCap.Cmp(gasFloor) < 0 {
+				transactor.GasFeeCap = gasFloor
+			}
+			if transactor.GasTipCap == nil || transactor.GasTipCap.Cmp(gasFloor) < 0 {
+				transactor.GasTipCap = gasFloor
```
ChiTimesChi, line 468: We were previously using `1 Gwei` as minimum value for GasPrice/GasFeeCap (but not GasTipCap).

Here we are setting Gas tip to 10 Gwei for first-time txs

ethergo/submitter/submitter.go
```diff
@@ -346,119 +346,211 @@ func (t *txSubmitterImpl) txTypeForChain(chainID *big.Int) (txType uint8) {
...
+				transactor.GasFeeCap = gasFloor
+				span.AddEvent("gas fee cap is nil; setting to gas floor")
+			} else {
+				transactor.GasFeeCap = gas.BumpByPercent(transactor.GasFeeCap, bumpPct)
+			}
+			if transactor.GasTipCap == nil {
+				transactor.GasTipCap = gasFloor
+				span.AddEvent("gas tip cap is nil; setting to gas floor")
+			} else {
+				transactor.GasTipCap = gas.BumpByPercent(transactor.GasTipCap, bumpPct)
```
ChiTimesChi, line 457: Do we need to check `transactor.GasTipCap` for nil here?

ethergo/submitter/submitter.go
```diff
@@ -346,119 +346,211 @@ func (t *txSubmitterImpl) txTypeForChain(chainID *big.Int) (txType uint8) {
...
-		if transactor.GasFeeCap != nil && maxPrice.Cmp(transactor.GasFeeCap) < 0 {
-			transactor.GasFeeCap = maxPrice
+	} else {
+		if prevDynamic {
+			transactor.GasPrice = core.CopyBigInt(prevTx.GasFeeCap())
+		} else {
+			transactor.GasPrice = core.CopyBigInt(prevTx.GasPrice())
 		}
+	}
+}
```
ChiTimesChi, line 424: Why aren't we setting the bumped values here? For me it makes sense to have "populate from old" and "bump by percentage" in the same function, as these always go together and refer to the same thing

ethergo/submitter/config/config.go
dwasse: This makes sense- I'm confused of the intention here though in the base PR?

https://github.com/synapsecns/sanguine/pull/2432/files#diff-9cb5c46a90d8a5d8f06ab1c53594fbc934b93eaf5131c538424f0c4666f94f49R429

To me this seems like we will never set gas price below 10 gwei if no `prevTx` is supplied

ethergo/submitter/submitter.go
```diff
@@ -346,119 +346,211 @@ func (t *txSubmitterImpl) txTypeForChain(chainID *big.Int) (txType uint8) {
...
+func (t *txSubmitterImpl) applyGasFromOracle(ctx context.Context, transactor *bind.TransactOpts, client client.EVM, useDynamic bool) (err error) {
+	ctx, span := t.metrics.Tracer().Start(ctx, "submitter.applyGasFromOracle")
+
+	defer func() {
+		metrics.EndSpanWithErr(span, err)
+	}()
 
-		transactor.GasTipCap, err = client.SuggestGasTipCap(ctx)
+	if useDynamic {
+		suggestedGasFeeCap, err := client.SuggestGasPrice(ctx)
```
ChiTimesChi, line 486: I suggest we use some kind of multiplier here. `SuggestGasPrice` returns base fee from the latest block + suggestedGasTip iirc, so a slight increase in the base fee will lead to tx not being included (as usually the gas tip is a tiny fraction of GasFeeCap).

ethergo/submitter/config/config.go
aureliusbtc: The unit there is in wei

ethergo/submitter/submitter.go
dwasse: We check `if transactor.GasTipCap == nil {` just above :)

ethergo/submitter/submitter.go
ChiTimesChi: Something like 1.5x-2x should work w/o forcing to spend extra, as you always get charged with _actual base fee_ + gas tip (that is unaffected by multiplier)

ethergo/submitter/submitter.go
```diff
@@ -346,119 +346,211 @@ func (t *txSubmitterImpl) txTypeForChain(chainID *big.Int) (txType uint8) {
...
+		span.SetAttributes(attribute.String("max_price", bigPtrToString(maxPrice)))
+		metrics.EndSpanWithErr(span, err)
+	}()
+
+	if useDynamic {
+		if transactor.GasFeeCap.Cmp(maxPrice) > 0 {
+			return fmt.Errorf("gas fee cap %s exceeds max price %s", transactor.GasFeeCap, maxPrice)
 		}
-		gas.BumpGasFees(transactor, t.config.GetGasBumpPercentage(chainID), gasBlock.BaseFee, maxPrice)
+		// TODO: should maxPrice apply to GasTipCap as well?
```
ChiTimesChi, line 532: We just need to enforce the invariant that `GasTipCap <= GasFeeCap`, so that this TODO check could be avoided

ethergo/submitter/submitter.go
dwasse: Yeah this makes sense, will simplify the code a bit. Will rename `populateGasFromPrevTx` since it will also be doing the bumping

ethergo/submitter/config/config.go
ChiTimesChi: Let's introduce a min value for GasPrice/GasFeeCap (DefaultMinGasPrice) and a separate one for GasTipCap. The latter one is set to 10 Wei (aka just 10) in the base PR.

ethergo/submitter/submitter.go
ChiTimesChi: Sorry, that was supposed to be a comment for this line:
https://github.com/synapsecns/sanguine/blob/6c52b8b698982134f37797897eb9fb8e63577264/ethergo/submitter/submitter.go#L459-L461

ethergo/submitter/submitter.go
dwasse: Done here: https://github.com/synapsecns/sanguine/pull/2462/commits/d245a735186ad8f6d12aeba5950b98198c5ca2bf

ethergo/submitter/submitter.go
dwasse: Will circle back in this issue: https://github.com/synapsecns/sanguine/issues/2464

ethergo/submitter/submitter.go
dwasse: Done here: https://github.com/synapsecns/sanguine/pull/2462/commits/b4394d1df4d201917205b8bd64b9beb7ea127ec6

ChiTimesChi:
@coderabbitai review

ethergo/submitter/config/config.go
```diff
@@ -110,17 +110,17 @@ func (c *Config) GetMaxGasPrice(chainID int) (maxPrice *big.Int) {
 	return
 }
 
-// GetBaseGasPrice returns the maximum gas price to use for transactions.
-func (c *Config) GetBaseGasPrice(chainID int) (basePrice *big.Int) {
-	basePrice = c.BaseGasPrice
+// GetMinGasPrice returns the maximum gas price to use for transactions.
```
ChiTimesChi, line 113:
```suggestion
// GetMinGasPrice returns the minimum gas price to use for transactions.
```

PR #2523, opened 2024-04-22
Feat: interchain check conflicts

ChiTimesChi:
**Description**
Changes behavior of the main entrypoint for batch verification: `InterchainDB.checkBatchVerification(module, batch)`. It never reverts, and has two edge cases with designated return values:
- `UNVERIFIED = 0` is returned if the module hasn't verified any batch that has the same `batch.srcChainId` and `batch.dbNonce` combination (aka the _batch key_).
- `CONFLICT = type(uint256).max` is returned if the module has verified a different batch root for the same _batch key_.

And as previously, it returns the timestamp when the module verified this exact batch, if this has happened in the past.

Worth noting that `InterchainDB` enforces the invariant that the same module can not verify different `batchRoot` values for the same _batch key_. Therefore once the module verified _something_ for `(srcChainId, dbNonce)`, the return value for `checkBatchVerification(module, (srcChainId, dbNonce, *))` is forever static for any passed `batchRoot`:
- verification timestamp for the correct batch root,
- `CONFLICT` for anything else.

**Motivation**
This PR enables the implementation of the Guard service as a standalone module. An additional check in the `InterchainClientV1` contract could be performed that the Guard module hasn't verified a conflicting value. This allows a straightforward Guard service implementation:
- Monitor the verified batches for a set of modules.
- Check the bacthes against the canonical state of the remote chain.
- In the event of a mismatch Guard should post the correct batch.

## Summary by CodeRabbit

- **New Features**
  - Enhanced the verification process for batch transactions in blockchain intercommunication, improving security and reliability.
  - Introduced new error handling for batch conflicts to aid in more precise debugging and conflict resolution.

- **Bug Fixes**
  - Updated verification logic to accurately reflect unverified states or conflicts, ensuring more robust transaction handling.

- **Documentation**
  - Added new constants and error messages to interface documentation to better guide developers in handling verification states.

### DIFF
--- a/packages/contracts-communication/contracts/InterchainClientV1.sol
+++ b/packages/contracts-communication/contracts/InterchainClientV1.sol
@@ -11,7 +11,7 @@ import {IInterchainDB} from "./interfaces/IInterchainDB.sol";
 
 import {AppConfigV1, AppConfigLib} from "./libs/AppConfig.sol";
 import {BatchingV1Lib} from "./libs/BatchingV1.sol";
-import {InterchainBatch} from "./libs/InterchainBatch.sol";
+import {InterchainBatch, InterchainBatchLib} from "./libs/InterchainBatch.sol";
 import {
     InterchainTransaction, InterchainTxDescriptor, InterchainTransactionLib
 } from "./libs/InterchainTransaction.sol";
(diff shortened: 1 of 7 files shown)


ChiTimesChi:
@coderabbitai review

aureliusbtc (approved):

PR #2534, opened 2024-04-29
feat: unify claiming in `SynapseModule` and `SynapseExecutionService`

ChiTimesChi:
**Description**
Since #2524 `SynapseExecutionService` is accumulating the paid execution fees. This PR introduces the claiming structure similar to one existing in `SynapseModule`, and does a step further to abstract it away from both contracts.

The claiming structure is:
- All native token balance is considered "accumulated fees" and is claimable to a governor-defined fee recipient address (which is the same as executor EOA for the service).
- Anyone can claim the fees, which will transfer the accumulated fees to the fee recipient.
- The governor could set up a fraction of fees that will go to the claimer (whoever triggered the claim).

This structure allows anyone to integrate a keeper bot that will periodically claim the fees, making both module and execution service self-sustainable, without having to transfer the collected gas fees per every occasion.


ChiTimesChi:
@coderabbitai review

aureliusbtc (approved):

PR #2607, opened 2024-05-10
feat(synapse-interface): refresh stale quotes

bigboydiamonds:
**Description**
Add hook to manage refreshing stale quotes based on `staleTimeout` in ms.

## Summary by CodeRabbit

- **New Features**
  - Added a timestamp attribute to certain data constants to enhance data structure.
  - Introduced a custom React hook for refreshing quotes when they become stale.
  - Implemented a new function to calculate time differences between two timestamps.

- **Bug Fixes**
  - Adjusted error logging behavior in the `formatAmount` function by disabling a console error output.

- **Refactor**
  - Updated quote management processes to include timestamp handling in various functions and components.

1a0ebf35886568fe41e34413017d80e888d6837c: [synapse-interface preview link ](https://sanguine-synapse-interface-dkdyquhll-synapsecns.vercel.app)
8ac4392f17c009ca2ee298436ceccb2d54db6ddc: [synapse-interface preview link ](https://sanguine-synapse-interface-25k4wmjcx-synapsecns.vercel.app)
a932e3c94d4a399aa326bb210b8e75fb08d9fe64: [synapse-interface preview link ](https://sanguine-synapse-interface-4hb7o3ogt-synapsecns.vercel.app)
56f21dfab051bc6c18fa6274e15f53a22cb40c88: [synapse-interface preview link ](https://sanguine-synapse-interface-6r3rahpvb-synapsecns.vercel.app)
924eae382dc8718a2491a84f8d92b370d6b08a0c: [synapse-interface preview link ](https://sanguine-synapse-interface-9ol9qz7un-synapsecns.vercel.app)
14527c1896ad0412601c008d9e7e49c81e121ff4: [synapse-interface preview link ](https://sanguine-synapse-interface-d7czlfj62-synapsecns.vercel.app)
252912fc2f7da8b7dc8f38ee361d205bc2bd1f16: [synapse-interface preview link ](https://sanguine-synapse-interface-inchrtavc-synapsecns.vercel.app)
cf34b600339a5799d2195c7b10209a69def32840: [synapse-interface preview link ](https://sanguine-synapse-interface-5zrt108fv-synapsecns.vercel.app)
16def59d163793f04be9814df7f6a54593ac700d: [synapse-interface preview link ](https://sanguine-synapse-interface-ocr3h2jmv-synapsecns.vercel.app)
4002ed7a3a05204ef20d02aff805b7365c7caa66: [synapse-interface preview link ](https://sanguine-synapse-interface-e4920to49-synapsecns.vercel.app)


packages/synapse-interface/utils/hooks/useStaleQuoteRefresher.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (isValidQuote && !isQuoteLoading && !isWalletPending) {
+      const timeDifference = calculateTimeBetween(currentTime, quoteTime)
+      const isStaleQuote = timeDifference >= staleTimeout
+      if (isStaleQuote) {
+        document.addEventListener('mousemove', refreshQuoteCallback, {
+          once: true,
+        })
+      }
+    }
+  }, [currentTime])
```
abtestingalpha, line 33: Do we need the other deps in here?:

```
[currentTime, isValidQuote, isQuoteLoading, isWalletPending, quoteTime, staleTimeout])
```

bigboydiamonds: `currentTime` is the only dep we would require to refresh a stale quote. Including the other deps may prematurely refresh the quote in this useEffect

abtestingalpha: `isValidQuote, isQuoteLoading, isWalletPending`: If I'm understanding this correctly, these flags determine if an event listener should be added? And we don't want to add them bc we don't want to re-attach if they change?

`quoteTime`: Is this the timestamp which staleness is calculated? Don't we need to include that bc if `quoteTime` changes, then the staleness logic needs to re-run?

`staleTimeout`: Is this the value that determines what's considered stale? So if it changes, the staleness logic would re-run here as well?

bigboydiamonds: Correct, `isValidQuote, isQuoteLoading, isWalletPending` are used to determine if an event listener should be attached, but we don't want to react to these props because: 
- If wallet action is pending, it would refresh quote in background, but does not actually update the quote used for the transaction (prevents User confusion with output amounts)
- Listening to `isQuoteLoading` could cause for infinite re-renders 
- `isValidQuote` may change frequently if the User is testing out various inputs (token / amount combos) and could effectively re-trigger a quote fetch before interval time has passed 

Including `quoteTime` would also cause infinite re-renders since the change in `quoteTime` would cause for the quote to refresh in the background, which then updates the `quoteTime` value again. 

Good catch on the `staleTimeout`! I originally expected for `staleTimeout` to be a static value when using the hook, but that's an improper assumption to make. Updated to include `staleTimeout` in the useEffect dependencies array in https://github.com/synapsecns/sanguine/pull/2607/commits/f8fadd6fc5f2cd75378403a7841e702864e29a2c

PR #2616, opened 2024-05-11
feat(widget): maintenance

bigboydiamonds:


packages/widget/src/utils/getSynapsePauseData.ts
```diff
@@ -0,0 +1,70 @@
...
+      const msPerHr = 1000 * 60 * 60
+      const timeElapsedInHrs = (currentTime - previousTime) / msPerHr
+
+      return timeElapsedInHrs < 1
+    } else {
+      return false
+    }
+  }
+
+  const isValid = checkIsDataValid()
```
abtestingalpha, line 63: Is this so it only updates the session store after an hour has passed? If it's to prevent http calls, don't think it's worth it since there's a condition where a user used the widget in the time between last use and when we push a chain pause. Fine to check each time.

bigboydiamonds: Thanks, updated to remove `checkIsDataValid()`, now checking every time in https://github.com/synapsecns/sanguine/commit/fc8c79c22d2be6c8d2555359fdb877e777a675aa

Added `isFetching` flag to prevent duplicate fetches on re-renders.

PR #2719, opened 2024-06-12
RFQ Relayer: speed improvements

dwasse:
## Summary by CodeRabbit

- **New Features**
  - Introduced parallel processing of quote requests with semaphore control for improved performance.
  - Added SQL statistics tracking to monitor database performance.

- **Improvements**
  - Enhanced error handling with retry mechanisms for certain contract calls.
  - Implemented mutex handling for better concurrency control.
  
- **Refactor**
  - Streamlined request processing to support both serial and parallel execution.


services/rfq/relayer/reldb/base/rebalance.go
```diff
@@ -125,3 +127,13 @@ func (s Store) GetRebalanceByID(ctx context.Context, rebalanceID string) (*reldb
 
 	return rebalance, nil
 }
+
+// GetDBStats gets the database stats.
+func (s Store) GetDBStats(ctx context.Context) (*sql.DBStats, error) {
```
trajan0x, line 132: I like this

services/rfq/relayer/service/handlers.go
```diff
@@ -201,20 +233,37 @@ func (q *QuoteRequestHandler) handleCommitPending(ctx context.Context, span trac
 		return nil
 	}
 
-	bs, err := q.Origin.Bridge.BridgeStatuses(&bind.CallOpts{Context: ctx}, request.TransactionID)
+	var bs uint8
+	call := func(ctx context.Context) error {
```
trajan0x, line 237: If we're going to be doing this all over the place, I'd recommend instead wrapping [ethergo/client](https://pkg.go.dev/github.com/synapsecns/sanguine/ethergo/client#DialBackend) and having a func called something like `DialRetryableBackend`

Alchemy has a good [guide](https://docs.alchemy.com/docs/how-to-implement-retries) on retries that might be good supplemental reading, but exponential retries w/ cap seems like the way to go.

I say this because there's nothing "special" about this call, these can all fail and we want to retry on *http errors*, not evm retry errors

services/rfq/relayer/service/handlers.go
```diff
@@ -81,11 +93,23 @@ func (r *Relayer) handleBridgeRequestedLog(parentCtx context.Context, req *fastb
...
 		return fmt.Errorf("could not get db: %w", err)
 	}
 
+	// immediately forward the request to handleSeen
+	span.AddEvent("forwarding to handleSeen")
+	qr, err := r.requestToHandler(ctx, dbReq)
+	if err != nil {
+		return fmt.Errorf("could not get quote request handler: %w", err)
+	}
+	err = qr.handleSeen(ctx, span, dbReq)
```
trajan0x, line 108: This method has two issues:

1. calling the method like this avoids the deadline middleware and the gas middleware
2. calling this method create's a race condition that can result in two relays being submitted at the same time.

WRT 1., just call qr.Handle() instead of handle Seen

WRT 2. As @aureliusbtc pointed out in slack, this creates a race condition where we can submit two "Relay"'s at the same time

I'd recommend creating a new middleware we apply to all functions that uses [stringmutex](https://pkg.go.dev/github.com/synapsecns/sanguine/core/mapmutex#StringMapMutex) to lock each tx so it won't get processed more than once

services/rfq/relayer/service/relayer.go
```diff
@@ -297,35 +299,62 @@ func (r *Relayer) startCCTPRelayer(ctx context.Context) (err error) {
...
+
+	g, ctx := errgroup.WithContext(ctx)
 	// Obviously, these are only seen.
-	for _, request := range requests {
-		// if deadline < now
-		if request.Transaction.Deadline.Cmp(big.NewInt(time.Now().Unix())) < 0 && request.Status.Int() < reldb.RelayCompleted.Int() {
-			err = r.db.UpdateQuoteRequestStatus(ctx, request.TransactionID, reldb.DeadlineExceeded)
+	for _, req := range requests {
+		request := req // capture func literal
+		g.Go(func() error {
```
trajan0x, line 318: Think it's a very bad idea not to have *any kind of* limit on number of go routines we spawn, seven if we keep it at 50.

This doesn't have to be a lot of code, just sem.Acquire and sem.release like we do in ethergo/chain/client

trajan0x (changes requested):

services/rfq/relayer/service/handlers.go
dwasse: Hmm, if we implement the mutex as a middleware don't we lose all the concurrency gains? Since each request handling will have to be in serial?

What if we just put a lock around submitter to make sure SubmitTransaction() can only be called once concurrently, while the rest of the request handling routine can be done in parallel?

Agree that we can also separate the processDB() loops into pre and post-relay txs

PR #2765, opened 2024-06-21
[DO NOT CLOSE] [DO NOT MERGE] Ponder/message bus

ChiTimesChi:
**Description**
Replaces #2514

### DIFF
--- a/packages/explorer-message-bus-api/.env.example
+++ b/packages/explorer-message-bus-api/.env.example
@@ -1,0 +1,16 @@
+# Change to "deploy" once 2024 events are synced
+# START_DATE=deploy
+START_DATE=2024-01-01
+
+# DFK
+PONDER_RPC_URL_53935=https://subnets.avax.network/defi-kingdoms/dfk-chain/rpc
+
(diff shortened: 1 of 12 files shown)


trajan0x:
@ChiTimesChi given that we're using this temporarily, any reason not to merge? Can nuke later

abtestingalpha:
@trajan0x @ChiTimesChi 

1. Why we can't merge yet: This branch is currently breaking the `synapse-interface` frontend build. Looking at CF logs, CF is unable to download Node 18.18. Rebuilds don't fix this issue. Unclear why this is happening, but guessing there's a conflict related to `sanguine` packages and ponder (which recommends `pnpm` and a different version of `viem`).
2. Since temporary, does this need to live in `sanguine`? If not, we can move it to its own repo with ponder-recommended package settings (`pnpm`) to avoid all these conflicts. And then point the railway build to that repo.

ChiTimesChi:
This package isn't actually using anything from the monorepo afaik, so could totally live as a separate repo.

PR #2787, opened 2024-06-25
RFQ Relayer: restrict state transitions

dwasse:


services/rfq/relayer/reldb/base/quote.go
```diff
@@ -89,7 +93,19 @@ func (s Store) GetQuoteResultsByStatus(ctx context.Context, matchStatuses ...rel
...
+func (s Store) UpdateQuoteRequestStatus(ctx context.Context, id [32]byte, status reldb.QuoteRequestStatus, prevStatus *reldb.QuoteRequestStatus) error {
+	if prevStatus == nil {
+		req, err := s.GetQuoteRequestByID(ctx, id)
+		if err != nil {
+			return fmt.Errorf("could not get quote: %w", err)
+		}
+		prevStatus = &req.Status
+	}
+	if !isValidStateTransition(*prevStatus, status) {
+		logger.Warnf("invalid state transition from %s to %s [txid=%s]", *prevStatus, status, hexutil.Encode(id[:]))
```
dwasse, line 105: Chose not to return error here since we presumably don't want to panic on this, and checking `errors.Is()` on the result introduces a bit of code bloat

core/mapmutex/mapmutex.go
```diff
@@ -82,6 +84,18 @@ func (m *untypedMapMutexImpl) TryLock(key interface{}) (Unlocker, bool) {
 	return nil, false
 }
 
+// Keys returns all keys in the map.
+func (m *untypedMapMutexImpl) Keys() []interface{} {
+	// note that we don't lock here in case the map is already locked.
```
trajan0x, line 89: @dwasse yeah this is just wrong imo. 

This is definitely going to nil pointer- if it was only used for tests ifne but this method is unsafe b/c if could lead to iterating over a non-existent slice element

The lock here isn't global anyway and won't block anything else for long at all. The whole point of mapmutex is the Locks are not global even - the lock here is internal only

trajan0x (changes requested):

core/mapmutex/mapmutex.go
```diff
@@ -2,8 +2,9 @@ package mapmutex
 
 import (
 	"fmt"
-	"github.com/LK4d4/trylock"
 	"sync"
+
+	"github.com/LK4d4/trylock"
```
trajan0x, line 7: Think we should remove this dep actually?

See: LK4D4/trylock#3

trajan0x (changes requested):

core/mapmutex/mapmutex.go
trajan0x: okay, i've decided we should address this in a different pr (only bc this requires make tidy)

tracking in #2814

core/mapmutex/mapmutex_test.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	. "github.com/stretchr/testify/assert"
 	"github.com/synapsecns/sanguine/core/mapmutex"
+	"gotest.tools/assert"
```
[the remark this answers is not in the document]
trajan0x: @dwasse also a good take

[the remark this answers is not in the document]
trajan0x: fixed in e867a9b1d

services/rfq/relayer/reldb/base/quote.go
trajan0x: Completely understand the hesitation to do this, but in this case I'll be renaming the method to `UpdateQuoteRequestStatusIFValid` or somethign.

trajan0x (changes requested):

core/mapmutex/mapmutex.go
trajan0x: fixed in 0d573f27371e39f063829348bf239f7fba9916b2

Issue #2750, opened 2024-06-20
[Submitter]: Cap TX count in chainConfirmQueue


trajan0x:
see thhis single batch w/ 9,800 receipt requests:

<img width="1687" alt="image" src="https://github.com/synapsecns/sanguine/assets/83933037/af4e2fdf-a36b-4945-9b8a-9c51e704f0f2">

PR #2789, opened 2024-06-25
Submitter: cap transactions in chainConfirmQueue

dwasse:
Addresses #2750 

## Summary by CodeRabbit

- **New Features**
	- Added functionality to limit the number of transactions per chain during sorting, enhancing transaction management.

- **Bug Fixes**
	- Updated test cases to ensure accurate sorting behavior with the new transaction cap.


trajan0x:
Edget case: this will break if there's more than 100 atetmpts for a tx

PR #2896, opened 2024-07-19
feat(sdk): uuid `BridgeQuote.id`

bigboydiamonds:


ChiTimesChi (commented):
This LGTM. Tried to think of a scenario where this could be a breaking change for an external integration, don't think so, right?

bigboydiamonds:
> This LGTM. Tried to think of a scenario where this could be a breaking change for an external integration, don't think so, right?

Agreed, don't believe this would be a breaking change for any external integrations.

trajan0x (approved):

PR #2917, opened 2024-07-25
feat(rfq-relayer): add MaxBalance param

dwasse:
## Summary by CodeRabbit

- **New Features**
	- Introduced minimum and maximum balance configuration for token quoting, allowing for more precise management of available funds.
	- Added functionality to retrieve minimum and maximum balances for specific tokens based on their configuration.
  
- **Bug Fixes**
	- Enhanced quoting logic to ensure that minimum and maximum balances are preserved when calculating quote amounts.
  
- **Tests**
	- Updated test cases to account for new minimum and maximum balance parameters, ensuring accurate validation of quoting behavior.


services/rfq/relayer/quoter/quoter.go
```diff
@@ -519,13 +519,17 @@ func (m *Manager) getOriginAmount(parentCtx context.Context, origin, dest int, a
 		quoteAmount = minQuoteAmount
 	}
 
-	// Finally, clip the quoteAmount by the balance
-	if quoteAmount.Cmp(balance) > 0 {
-		span.AddEvent("quote amount greater than balance", trace.WithAttributes(
+	// Finally, clip the quoteAmount by the minimum balance
+	minBalance := m.config.GetMinBalance(dest, address)
+	quotableBalance := new(big.Int).Sub(balance, minBalance)
```
ChiTimesChi, line 524: Hm, what if our balance is lower than minBalance?

services/rfq/relayer/quoter/quoter.go
```diff
@@ -519,17 +550,32 @@ func (m *Manager) getOriginAmount(parentCtx context.Context, origin, dest int, a
 		quoteAmount = minQuoteAmount
 	}
 
-	// Finally, clip the quoteAmount by the balance
-	if quoteAmount.Cmp(balance) > 0 {
-		span.AddEvent("quote amount greater than balance", trace.WithAttributes(
+	// Clip the quoteAmount by the max origin balance
+	maxBalance := m.config.GetMaxBalance(input.OriginChainID, input.OriginTokenAddr)
+	if input.OriginBalance != nil && maxBalance.Cmp(big.NewInt(0)) > 0 {
+		quotableBalance := new(big.Int).Sub(maxBalance, input.OriginBalance)
```
ChiTimesChi, line 556: I think we need to check if `quotableBalance < 0` here, as we could assign a negative value to `quoteAmount` below should the balance exceed maximum.

services/rfq/relayer/relconfig/getters.go
```diff
@@ -298,6 +298,43 @@ func (c Config) GetQuoteOffsetBps(chainID int, tokenName string, isOrigin bool)
 	return offset, nil
 }
 
+const defaultMaxBalance = 0
```
ChiTimesChi, line 301: Should we use something that resembles "positive infinity" here instead as a default value? This way we could actually use maxBalance = 0 to effectively stop quoting "to chain", while still quoting the other direction if that makes sense.

services/rfq/relayer/quoter/quoter.go
```diff
@@ -519,17 +550,32 @@ func (m *Manager) getOriginAmount(parentCtx context.Context, origin, dest int, a
 		quoteAmount = minQuoteAmount
 	}
 
-	// Finally, clip the quoteAmount by the balance
-	if quoteAmount.Cmp(balance) > 0 {
-		span.AddEvent("quote amount greater than balance", trace.WithAttributes(
+	// Clip the quoteAmount by the max origin balance
+	maxBalance := m.config.GetMaxBalance(input.OriginChainID, input.OriginTokenAddr)
+	if input.OriginBalance != nil && maxBalance.Cmp(big.NewInt(0)) > 0 {
```
ChiTimesChi, line 555: See comment above re: default value for `maxBalance` - this brings the question how do we want to treat `maxBalance == 0` case

services/rfq/relayer/relconfig/getters.go
dwasse: Good idea, will use `nil` for this

services/rfq/relayer/quoter/quoter.go
dwasse: Current thinking is to use `nil` for 'positive inf'- this way an unspecified `max_balance` param in the config will mean no max balance value, explicitly setting it to zero should result in all quotes to this chain having 0 quoteAmount

services/rfq/relayer/quoter/quoter.go
```diff
@@ -519,17 +550,39 @@ func (m *Manager) getOriginAmount(parentCtx context.Context, origin, dest int, a
...
+				attribute.String("max_balance", maxBalance.String()),
+				attribute.String("origin_balance", input.OriginBalance.String()),
+			))
+			quoteAmount = quotableBalance
+		}
+	}
+
+	// Finally, clip the quoteAmount by the dest balance
+	if quoteAmount.Cmp(input.DestBalance) > 0 {
+		span.AddEvent("quote amount greater than quotable balance", trace.WithAttributes(
```
ChiTimesChi, line 577: Should be "greater than destination balance"?

ChiTimesChi (approved):

PR #2919, opened 2024-07-25
feat(relconfig): decimals config validation

golangisfun123:
**Description**

**Additional context**

**Metadata**
- Fixes #[Link to Issue]

## Summary by CodeRabbit

- **New Features**
	- Enhanced command handling for improved interaction with the OmniRPC service.
	- Introduced a new method for validating token decimal values in the relayer configuration.

- **Bug Fixes**
	- Added validation checks to ensure token decimal configurations match actual blockchain values, preventing potential discrepancies.

- **Tests**
	- Implemented comprehensive tests for validating token decimal configurations, including checks for correct and incorrect settings across multiple tokens.


services/rfq/relayer/relconfig/config.go
```diff
@@ -228,5 +233,34 @@ func (c Config) Validate() (err error) {
...
+// ValidateTokenDecimals calls decimals() on the ERC20s to ensure that the decimals in the config match the actual token decimals.
+func (c Config) ValidateTokenDecimals(ctx context.Context, omniClient omniClient.RPCClient) (err error) {
+	for chainID, chainCfg := range c.Chains {
+		for tokenName, tokenCFG := range chainCfg.Tokens {
+			chainClient, err := omniClient.GetChainClient(ctx, chainID)
+			if err != nil {
+				return fmt.Errorf("could not get chain client for chain %d: %w", chainID, err)
+			}
+
+			ierc20, err := ierc20.NewIERC20(common.HexToAddress(tokenCFG.Address), chainClient)
```
ChiTimesChi, line 249: There's one edge case here when token address is `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE`. This is used to represent a native gas token (Ether) and therefore this address doesn't correspond to an actual smart contract.

In this edge case we should use 18 as `actualDecimals`

golangisfun123:
```go
			// Check if the token is the gas token. SHOULD BE 18.
			if tokenCFG.Address == chain.EthAddress.String() {
				if tokenCFG.Decimals != 18 {
					return fmt.Errorf("decimals mismatch for token %s on chain %d: expected 18, got %d", tokenName, chainID, tokenCFG.Decimals)
				}
				continue
			}
```

services/rfq/relayer/cmd/commands.go
```diff
@@ -36,9 +37,14 @@ var runCommand = &cli.Command{
 		if err != nil {
 			return fmt.Errorf("could not read config file: %w", err)
 		}
-
 		metricsProvider := metrics.Get()
 
+		omniClient := omniClient.NewOmnirpcClient(cfg.OmniRPCURL, metricsProvider, omniClient.WithCaptureReqRes())
+
+		if err = cfg.ValidateTokenDecimals(c.Context, omniClient); err != nil {
```
dwasse, line 44: IMO it's a bit unnatural to do config validation in the cmd, my recommendation would be to call `validateTokenDecimals()` within the existing `Validate()` function. Can check `if omniClient == nil` so that we handle cases where the caller may not have access to a client

dwasse (approved):

PR #2933, opened 2024-07-29
[SLT-149] feat(rfq): relay rate limiting

golangisfun123:
**Description**

**Additional context**

**Metadata**
- Fixes #[Link to Issue]

## Summary by CodeRabbit

- **New Features**
	- Introduced a rate-limiting mechanism for handling quote requests, improving request management.
	- Enhanced handling of token configurations in the relayer service.
	- Added configuration fields for `VolumeLimit`, enhancing transaction control.
	- Implemented new methods for retrieving limit confirmations and volume limits in the configuration.
	- Added a new configuration option, `volume_limit`, for better risk management in the relayer's YAML file.
	- Enhanced testing capabilities with a comprehensive test suite for the rate limiter.

- **Bug Fixes**
	- Improved logic for handling quote requests, ensuring better adherence to volume limits and rate limits.

1c4e49abffe5064fb43a2d021f749344a835221e: [docs preview link ](https://bridge-docs-r78ap7m19-synapsecns.vercel.app)
e2ec3da3b65adf9859cef331551da33591c2e16a: [docs preview link ](https://bridge-docs-9cptxhy8f-synapsecns.vercel.app)
1856513598484ee16cf25215053d08d1b722c8a0: [docs preview link ](https://bridge-docs-bgp42ievj-synapsecns.vercel.app)
35d5d24e778b0dc3f9984db75f5ad7f1620c5d37: [docs preview link ](https://bridge-docs-i2zcggyiz-synapsecns.vercel.app)
fecb6c552b2d566f837b1371e94db61a1be831cf: [docs preview link ](https://bridge-docs-9ft27a2n2-synapsecns.vercel.app)
3fec73a3058fc161810e9dec00198d6f0fc397f5: [docs preview link ](https://bridge-docs-d153skgui-synapsecns.vercel.app)
717f95fe90c9bf6ce38cf6dc5003897b4f97e32d: [docs preview link ](https://bridge-docs-3ny4rudsi-synapsecns.vercel.app)
8e25424d39e9b3ea20369e6a0bb5a0d329e2f692: [docs preview link ](https://bridge-docs-6mwf3hh8s-synapsecns.vercel.app)
491986546c0a4ff343d033c46c384e4076222244: [docs preview link ](https://bridge-docs-gksq5ocj2-synapsecns.vercel.app)
be538784ce077b7f7db2394f91b73c492d029b37: [docs preview link ](https://bridge-docs-a7tru2vtp-synapsecns.vercel.app)
e721d9caea1b09b93abe0733d219499bf1f16223: [docs preview link ](https://bridge-docs-kx0fh1u4z-synapsecns.vercel.app)
540148016b14cce510b1bd5f2d9bd1df9db9b764: [docs preview link ](https://bridge-docs-i7m69keb5-synapsecns.vercel.app)
109d11a1a344e64e655d04b2dc74e5fa99a9251b: [docs preview link ](https://bridge-docs-jvofzvzmw-synapsecns.vercel.app)
ecdd7e44a81c8814daab38146db3349296757956: [docs preview link ](https://bridge-docs-9t4tna4eu-synapsecns.vercel.app)
691d8406a4daf486964b07e9bc49d23bef111bf7: [docs preview link ](https://bridge-docs-rngqavkun-synapsecns.vercel.app)
a8b0fe2a59ef5c65a7e0c5f6cb75adb8c1bf6f5a: [docs preview link ](https://bridge-docs-hii8n0s1q-synapsecns.vercel.app)
92d7964f3cef005b18a8135c7b3041bf8dc98873: [docs preview link ](https://bridge-docs-qreoenjex-synapsecns.vercel.app)
c919b6d26ef178e6a041b13e320ce0842b951132: [docs preview link ](https://bridge-docs-koxd1tuwv-synapsecns.vercel.app)
62e8d15bde80990a0e1bb468e1c7b633127c70ed: [docs preview link ](https://bridge-docs-7msmj8xmb-synapsecns.vercel.app)
dc9b79590a7b6ad372d8e71b47483ae1507a0333: [docs preview link ](https://bridge-docs-2vbba5q9t-synapsecns.vercel.app)
ee0d5cf83c3eeb056c83d50a3c8631c04c981e21: [docs preview link ](https://bridge-docs-d41600frp-synapsecns.vercel.app)
9a7f39eb09e01feff3f8fc693c98ae3b885e28a1: [docs preview link ](https://bridge-docs-k6le17s56-synapsecns.vercel.app)
57d3fb3f72796e9e6b33cabaf6d62cd3961063f9: [docs preview link ](https://bridge-docs-l24ehk3e2-synapsecns.vercel.app)
3fdaf68198e1c8143ba05555d5910c5a4c860f19: [docs preview link ](https://bridge-docs-7fqcdu75j-synapsecns.vercel.app)
74dd575732cf62f21667db804dbc24ceda5f6782: [docs preview link ](https://bridge-docs-h5fmfx3x9-synapsecns.vercel.app)
397c378bfbb56f199aa1668fd77880592623e030: [docs preview link ](https://bridge-docs-2xkaln0ww-synapsecns.vercel.app)
95c3e4d504770f790a90c9d987f03a47efc40e81: [docs preview link ](https://bridge-docs-a9cosz54p-synapsecns.vercel.app)
1973d848fb752fb678068456abd03f7898f8433a: [docs preview link ](https://bridge-docs-d6o5y8t2j-synapsecns.vercel.app)
6a138b4d8fcd24e189304e2b1c892e6d78203cfe: [docs preview link ](https://bridge-docs-n0wqihg82-synapsecns.vercel.app)
e252d2eb349b58a8a70947da5fe4b2238ff999cd: [docs preview link ](https://bridge-docs-4lr8kt1gd-synapsecns.vercel.app)
0cbf883c3bde6707d08b8c0728f10ee05db3842e: [docs preview link ](https://bridge-docs-2przf4kav-synapsecns.vercel.app)


services/rfq/relayer/service/statushandler.go
```diff
@@ -224,6 +238,103 @@ func (q *QuoteRequestHandler) shouldCheckClaim(request reldb.QuoteRequest) bool
...
+// Case 2: We should also not relay the request if the cumulative relayed amount over the last blockWindowSize blocks
+// exceeds the volume limit AND we have not confirmed.
+func (q *QuoteRequestHandler) canRelayBasedOnVolumeAndConfirmations(
+	request reldb.QuoteRequest,
+	currentBlockNumber uint64,
+	volumeLimit float64,
+) (bool, error) {
+	// Case 1: Singular RFQ over volumeLimit and inadequate confirmations
+
+	priceOfOriginToken, err := q.getTokenPrice(context.Background(), request)
```
dwasse, line 253: Let's pass `ctx` in as an argument to `canRelayBasedOnVolumeAndConfirmations`. In general `context.Background()` should only really be used when the process is initially spun up

services/rfq/relayer/service/statushandler.go
```diff
@@ -224,6 +238,103 @@ func (q *QuoteRequestHandler) shouldCheckClaim(request reldb.QuoteRequest) bool
...
+	if err != nil {
+		return false, fmt.Errorf("could not get price: %w", err)
+	}
+
+	numOfConfirmations := currentBlockNumber - request.BlockNumber
+
+	if priceOfOriginToken > volumeLimit && numOfConfirmations < 1 {
```
dwasse, line 260: We should take the size of the transaction into account here right? If you just use `priceOfOriginToken` we are assuming size of 1 (since we are comparing it to volume)?

services/rfq/relayer/service/statushandler.go
```diff
@@ -224,6 +238,103 @@ func (q *QuoteRequestHandler) shouldCheckClaim(request reldb.QuoteRequest) bool
...
+	priceOfOriginToken, err := q.getTokenPrice(ctx, request)
+	if err != nil {
+		return fmt.Errorf("could not get price: %w", err)
+	}
+
+	// If we have some room, add the RFQ to the cache.
+	cacheLength := q.rfqCache.Len()
+	if cacheLength < q.blockWindowSize {
+		prev := q.rfqCache.GetOrDefault(request.BlockNumber, 0)
+		q.rfqCache.Set(request.BlockNumber, prev+priceOfOriginToken)
```
dwasse, line 294: `rfqCache` is meant to store USD volume right? in this case we should be inserting `priceOfOriginToken * request.OriginAmount` (where origin amount is adjusted for decimals)

services/rfq/relayer/service/handlers.go
```diff
@@ -186,6 +186,22 @@ func (q *QuoteRequestHandler) handleSeen(ctx context.Context, span trace.Span, r
...
 
+	latestBlock := q.Origin.LatestBlock()
+
+	canRelay, err := q.canRelayBasedOnVolumeAndConfirmations(request, latestBlock, q.volumeLimit)
+	if err != nil {
+		span.AddEvent("could not determine if can relay")
+		return fmt.Errorf("could not determine if can relay: %w", err)
+	}
+	if !canRelay {
+		err = q.db.UpdateQuoteRequestStatus(ctx, request.TransactionID, reldb.CommittedPending, &request.Status)
```
dwasse, line 197: IMO we shouldn't update to CommittedPending here because we don't want to allow the balance to be consumed by another request until this one is ready to relay

golangisfun123: at this point we de facto committed confirmed this, we just cant act on it yet? it follow similar order to 

```
func (q *QuoteRequestHandler) handleCommitPending(ctx context.Context, span trace.Span, request reldb.QuoteRequest) (err error) {
	earliestConfirmBlock := request.BlockNumber + q.Origin.Confirmations

	latestBlock := q.Origin.LatestBlock()
	shouldContinue := latestBlock >= earliestConfirmBlock
	span.AddEvent("pending_check", trace.WithAttributes(
		attribute.Int("latest_block", int(latestBlock)),
		attribute.Int("earliest_confirm_block", int(earliestConfirmBlock)),
		attribute.Bool("should_continue", shouldContinue),
	))
	...
```

services/rfq/relayer/limiter/limiter.go
```diff
@@ -0,0 +1,116 @@
...
+	tokenPrice, err := l.getUSDAmountOfToken(ctx, l.quoter, request)
+	if err != nil {
+		return false, fmt.Errorf("could not get USD amount of token: %w", err)
+	}
+	return tokenPrice <= l.cfg.VolumeLimit, nil
+}
+
+func (l *limiterImpl) getUSDAmountOfToken(
+	ctx context.Context,
+	q quoter.Quoter,
```
dwasse, line 99: `limiterImpl` already has access to a quoter right? So no need to pass it as an arg here

services/rfq/relayer/limiter/limiter.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/synapsecns/sanguine/services/rfq/relayer/reldb"
+	"go.uber.org/ratelimit"
+)
+
+// Limiter is the interface for rate limiting RFQs.
+type Limiter interface {
+	// IsAllowed returns true if the request is allowed, false otherwise.
+	IsAllowed(ctx context.Context, request reldb.QuoteRequest) (bool, error)
+	// Take blocks until a token is available.
+	Take() time.Time
```
dwasse, line 23: Unused atm? I assume you plan to use this in the stateful version?

services/rfq/relayer/limiter/limiter.go
```diff
@@ -0,0 +1,116 @@
...
+			break
+		}
+	}
+
+	price, err := q.GetPrice(ctx, tokenName)
+	if err != nil {
+		return 0, fmt.Errorf("could not get price: %w", err)
+	}
+
+	return price * float64(request.Transaction.OriginAmount.Int64()), nil
```
dwasse, line 115: A couple things to consider here:

- I believe `request.Transaction.OriginAmount` corresponds to the onchain amount (that is, it has no decimals). We should probably use `big.Int`math here to preserve precision
- Best way to handle this IMO is have the decimal conversion in the getter, so that the config specifies a value in human-readable units, but internally you deal with the scaled-up big.Int value. See `GetMaxBalance()` and `GetMinQuoteAmount()` in `getters.go` for reference here

services/rfq/relayer/limiter/limiter.go
```diff
@@ -0,0 +1,116 @@
...
+	tokenNames map[string]relconfig.TokenConfig
+}
+
+// NewRateLimiter creates a new Limiter.
+// TODO: implement the sliding window: queue up requests and process them in order if cumulative volume is above limit.
+func NewRateLimiter(
+	cfg relconfig.Config,
+	q quoter.Quoter,
+	client client.EVM,
+	metricHandler metrics.Handler,
```
dwasse, line 42: Let's definitely add some traces to functions here, particularly when we make the final `IsAllowed()` decision

services/rfq/relayer/relconfig/config.go
```diff
@@ -59,6 +59,10 @@ type Config struct {
 	UseEmbeddedGuard bool `yaml:"enable_guard"`
 	// SubmitSingleQuotes enables submitting single quotes.
 	SubmitSingleQuotes bool `yaml:"submit_single_quotes"`
+	// MaxRFQSize is the amount of RFQs we are allowed to process
+	MaxRFQSize int `yaml:"rfq_size"`
```
dwasse, line 63: Is this supposed to be `MinConfirmations`? Naming is a bit confusing to me here. I would recommend a more detailed explanation (either here or in the actual `limiter.go` file) with how exactly these parameters are expected to filter a given tx

dwasse:
Left a few comments but looking good so far, I think OK to stage once remaining comments are addressed

services/rfq/relayer/relconfig/getters.go
```diff
@@ -691,3 +692,17 @@ func (c Config) GetQuoteSubmissionTimeout() time.Duration {
 	}
 	return timeout
 }
+
+const defaultBlockWindow = 5
+
+const defaultVolumeLimit = 10_000
+
+// GetVolumeLimit returns the volume limit for the relayer.
+func (c Config) GetVolumeLimit() float64 {
```
dwasse, line 701: Is VolumeLimit meant to be in human-readable units (with decimals) or no? It seems like it's used as non-human-readable here, but in that case we should use `big.Int` instead of `float64`

Though I think less error prone to make this a human-readable value with a getter that does the decimal conversion

See how we do it here: https://github.com/synapsecns/sanguine/blob/master/services/rfq/relayer/relconfig/getters.go#L423

contrib/screener-api/go.mod
```diff
@@ -39,7 +39,6 @@ require (
 require (
 	dario.cat/mergo v1.0.0 // indirect
 	github.com/Azure/go-ansiterm v0.0.0-20210617225240-d185dfc1b5a1 // indirect
-	github.com/BurntSushi/toml v1.4.0 // indirect
```
trajan0x, line 42: going forward if you're seeing these kinds of things, you can run [`make tidy`](https://github.com/synapsecns/sanguine/blob/master/make/repo.Makefile#L29-L31) from the repo root.

I've done this for you in aba64a79c79d30a8502a1213274d893f75b0124d

going forward the `needs-go-generate`... label is a hint you need to do this particularly this one: https://github.com/synapsecns/sanguine/pull/2933#event-13678590298) the other was caused by https://github.com/synapsecns/sanguine/pull/2933#issuecomment-2278707844

trajan0x:
needs-go-generate-promexporter is a result of dfk models being out of date
image

fixed in #3004 merged into master in 2842a3adea8317f2448862b2902c6d57aba9c73b and merged back in into this pr in ce3a49904ac41cdb7b57ca8cf87c8ce0fcccedc7

services/rfq/relayer/limiter/limiter.go
```diff
@@ -0,0 +1,162 @@
...
+		metrics:    metricHandler,
+		quoter:     q,
+		cfg:        cfg,
+		tokenNames: tokens,
+	}
+}
+
+// IsAllowed returns true if the request is allowed, false otherwise.
+func (l *limiterImpl) IsAllowed(ctx context.Context, request reldb.QuoteRequest) (_ bool, err error) {
+	ctx, span := l.metrics.Tracer().Start(ctx, "limiter.IsAllowed", trace.WithAttributes(
```
trajan0x, line 54: I think this is overkill, but that's more of a nit.

My bigger ask is if we're going to do this can we have a helper `quoteRequestToAttributes` func in some kinda utils file?

I have something similiar [here](https://github.com/synapsecns/sanguine/blob/085a75892aaeeaa2d30f164137e196855cc3254d/ethergo/submitter/util.go#L52-L87) for reference

services/rfq/relayer/limiter/limiter.go
```diff
@@ -0,0 +1,160 @@
...
+// TODO: implement the sliding window: queue up requests and process them in order if cumulative volume is above limit.
+func NewRateLimiter(
+	cfg relconfig.Config,
+	q quoter.Quoter,
+	client client.EVM,
+	metricHandler metrics.Handler,
+	tokens map[string]relconfig.TokenConfig,
+) Limiter {
+	return &limiterImpl{
+		// Limiter:    ratelimit.New(cfg.MaxRFQSize),
```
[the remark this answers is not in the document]
trajan0x: nuke this line @golangisfun123

services/rfq/relayer/limiter/limiter.go
```diff
@@ -0,0 +1,162 @@
...
+	return withinSize, nil
+}
+
+func (l *limiterImpl) hasEnoughConfirmations(ctx context.Context, request reldb.QuoteRequest) (_ bool, err error) {
+	ctx, span := l.metrics.Tracer().Start(ctx, "limiter.hasEnoughConfirmations")
+	defer func() {
+		metrics.EndSpanWithErr(span, err)
+	}()
+
+	currentBlockNumber, err := l.client.BlockNumber(ctx)
```
trajan0x, line 96: How often is this called? Honestly we could just pull this from listener to save some rpc calls. Depends how performance conciensce we want to be here

Add the very least I'd add a TODO about optimization and/or some kinda caching here

services/rfq/relayer/limiter/limiter.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return false, fmt.Errorf("could not get block number: %w", err)
+	}
+
+	requiredConfirmations, err := l.cfg.GetConfirmations(int(request.Transaction.OriginChainId))
```
trajan0x, line 101: I don't think Confirmations here means what you think it does. 

https://github.com/synapsecns/sanguine/blame/2842a3adea8317f2448862b2902c6d57aba9c73b/services/rfq/relayer/relconfig/getters.go#L205 

This refers to number of confirmations for *any* tx, not foro txes over rate limit

services/rfq/relayer/limiter/limiter.go
trajan0x: Seperately, let's fetch requiredConfirmations first and fi ti's 0 return true. No reason to execute the rest of this routine otherwise (or even start the span)

services/rfq/relayer/limiter/limiter_test.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"math/big"
+	"time"
+
+	"github.com/stretchr/testify/mock"
+	clientMocks "github.com/synapsecns/sanguine/ethergo/client/mocks"
+	"github.com/synapsecns/sanguine/services/rfq/contracts/fastbridge"
+	"github.com/synapsecns/sanguine/services/rfq/relayer/limiter"
+	"github.com/synapsecns/sanguine/services/rfq/relayer/quoter/mocks"
+		"github.com/synapsecns/sanguine/services/rfq/util"
```
trajan0x, line 12: indent. Whole pr needs to run `make lint`

services/rfq/relayer/service/export_test.go
```diff
@@ -1,6 +1,8 @@
 package service
 
-import "context"
```
trajan0x, line 3: once again make lint

services/rfq/relayer/service/statushandler.go
```diff
@@ -81,6 +98,10 @@ func (r *Relayer) requestToHandler(ctx context.Context, req reldb.QuoteRequest)
 		apiClient:           r.apiClient,
 		mutexMiddlewareFunc: r.mutexMiddleware,
 		handlerMtx:          r.handlerMtx,
+		volumeLimit:         r.cfg.GetVolumeLimit(int(origin.ChainID), req.Transaction.OriginToken),
```
trajan0x, line 101: this doesn't appear to be used anywhere 
image

services/rfq/relayer/relconfig/getters.go
```diff
@@ -808,3 +808,28 @@ func (c Config) GetQuoteSubmissionTimeout() time.Duration {
 	}
 	return timeout
 }
+
+var defaultVolumeLimit *big.Int
```
trajan0x, line 812: @golangisfun123  sorry is this supposed to be nil by default?

Right now this will error if unset here: https://github.com/synapsecns/sanguine/blob/aba64a79c79d30a8502a1213274d893f75b0124d/services/rfq/relayer/relconfig/getters.go#L818

It's not even clear how this is supposed to work. If we expect 0 by default, then all txes are over volume by default since this will always return false

https://github.com/synapsecns/sanguine/blob/aba64a79c79d30a8502a1213274d893f75b0124d/services/rfq/relayer/limiter/limiter.go#L127

services/rfq/relayer/service/handlers.go
```diff
@@ -186,6 +186,19 @@ func (q *QuoteRequestHandler) handleSeen(ctx context.Context, span trace.Span, r
 		return nil
 	}
 
+	allowed, err := q.limiter.IsAllowed(ctx, request)
```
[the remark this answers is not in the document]
trajan0x: can you elaborate?

trajan0x (changes requested):
Relayer docs need to be updated in this PR as well

specifically exaple config section [here](https://github.com/synapsecns/sanguine/blob/aba64a79c79d30a8502a1213274d893f75b0124d/docs/bridge/docs/rfq/Relayer/Relayer.md)

services/rfq/relayer/limiter/limiter.go
golangisfun123: how does contract listener save RPC calls? 
i agree that client is overkill and probably is not the best choice for this though

services/rfq/relayer/relconfig/getters.go
golangisfun123: fixed, added a check. 0 is a no-limit value. 
`var defaultVolumeLimit *big.Int = common.Big0`

```
	volumeLimit := l.cfg.GetVolumeLimit(int(request.Transaction.OriginChainId), request.Transaction.OriginToken)
	// There is no limit.
	if volumeLimit.Cmp(common.Big0) == 0 {
		return true, nil
	}
```

golangisfun123: we can also do a nil check instead, what do you think is better?

services/rfq/relayer/limiter/limiter.go
golangisfun123: should I add another field to the config specifically for rate limit confirmations then?

services/rfq/relayer/relconfig/getters.go
trajan0x: I'd do -1 actually. We might want to do 0 to disable, so I'd default to -1 

btw, b/c big ints are pointers I generally discourage the use of common.Big0 type stuff since these  can be mutated as a side effect, especially by a bug in another lib dependent on 

```go
common.Big0 = big.New(5)
```

we also copy any globals to prevent accidental mutation.

services/rfq/relayer/service/handlers.go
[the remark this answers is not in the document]
trajan0x: @greptile-apps??

services/rfq/relayer/limiter/limiter.go
trajan0x: listener needs to track current blocknumber anyway and exposes it through 

https://pkg.go.dev/github.com/synapsecns/sanguine/ethergo@v1.1.0/listener#pkg-types

so you could pass in the listener instead as `type LatestBlockFetcher = interface {
    LatestBlock() int
}`

trajan0x (changes requested):

trajan0x:
@golangisfun123 don't forget to update docs

services/rfq/relayer/limiter/limiter.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func (l *limiterImpl) withinSizeLimit(ctx context.Context, request *reldb.QuoteRequest) (bool, error) {
+	volumeLimit := l.cfg.GetVolumeLimit(int(request.Transaction.OriginChainId), request.Transaction.OriginToken)
+	// There is no limit.
+	if volumeLimit.Cmp(big.NewInt(-1)) == 0 {
+		return true, nil
+	}
+
+	tokenPrice, err := l.getUSDAmountOfToken(ctx, request)
```
dwasse, line 121: Let's give this a more accurate name, like USDVolume or something like that

dwasse (approved):

PR #2994, opened 2024-08-08
feat(rfq-relayer): add mutex on committable balance consumption

dwasse:
## Summary by CodeRabbit

- **New Features**
  - Introduced a new locking mechanism for balance management to enhance concurrency control.
  - Added a mutex for balance requests in the relayer to synchronize access.

- **Improvements**
  - Enhanced the `QuoteRequestHandler` with a mutex for thread-safe balance operations.
  - Improved documentation for mutex usage in the relayer struct for clarity.

### DIFF
--- a/services/rfq/relayer/service/handlers.go
+++ b/services/rfq/relayer/service/handlers.go
@@ -498,12 +498,23 @@ func (q *QuoteRequestHandler) handleProofPosted(ctx context.Context, span trace.
 // Error Handlers Only from this point below.
 //
 // handleNotEnoughInventory handles the not enough inventory status.
-func (q *QuoteRequestHandler) handleNotEnoughInventory(ctx context.Context, _ trace.Span, request reldb.QuoteRequest) (err error) {
+func (q *QuoteRequestHandler) handleNotEnoughInventory(ctx context.Context, span trace.Span, request reldb.QuoteRequest) (err error) {
+	// acquire balance lock
+	key := getBalanceMtxKey(q.Dest.ChainID, request.Transaction.DestToken)
+	span.SetAttributes(attribute.String("balance_lock_key", key))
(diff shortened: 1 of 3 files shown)


services/rfq/relayer/service/handlers.go
ChiTimesChi, line 1: @dwasse I believe we also need to acquire the balance mutex in `handleNotEnoughInventory`, where we also do `GetCommittableBalance -> assign CommittedPending`

services/rfq/relayer/service/handlers.go
```diff
@@ -162,6 +162,17 @@ func (q *QuoteRequestHandler) handleSeen(ctx context.Context, span trace.Span, r
...
+	// lock the consumed balance
+	key := getBalanceMtxKey(q.Dest.ChainID, request.Transaction.DestToken)
+	span.SetAttributes(attribute.String("balance_lock_key", key))
+	unlocker, ok := q.balanceMtx.TryLock(key)
+	if !ok {
+		// balance is locked due to concurrent request, try again later
+		span.SetAttributes(attribute.Bool("locked", true))
+		return nil
+	}
+	defer unlocker.Unlock()
```
ChiTimesChi, line 174: In the case of forwarding, won't the lock be released only after the full cycle goes through? We could safely unlock once we update status to `CommittedPending` (aka spend the destination commitable balance)

ChiTimesChi: @dwasse by "full cycle" I mean the full status path `Seen -> CommitPending -> CommittedConfirmed`. I realised that it's not actually full, but still if that's the case we're keeping the lock longer than we should've

dwasse: Yeah that makes sense, would still like to use `defer` to make this error resistant but can just add a helper function here

PR #3003, opened 2024-08-09
refactor(rfq-relayer): rebalancing with multiple tokens

dwasse:
## Summary by CodeRabbit

- **New Features**
  - Simplified rebalance method now handles multiple tokens internally without requiring specific token parameters.
  - Introduced new methods for improved rebalance management, allowing for better performance and flexibility.

- **Improvements**
  - Enhanced clarity and efficiency in rebalance logic through centralized handling and more modular design.
  - Improved error handling and observability during rebalance execution.
  - Expanded test coverage for rebalance functionality, validating various scenarios and configurations.
  - Added new fields to the rebalance model for better tracking of token information.
  - Added a new linter to improve code quality checks by detecting duplication.
  - Updated handling of relay completions to focus on the destination chain.


services/rfq/relayer/inventory/manager.go
```diff
@@ -458,50 +454,53 @@ func (i *inventoryManagerImpl) HasSufficientGas(parentCtx context.Context, chain
...
-	ctx, span := i.handler.Tracer().Start(parentCtx, "Rebalance", trace.WithAttributes(
-		attribute.Int(metrics.ChainID, chainID),
-		attribute.String("token", token.Hex()),
+	return nil
+}
+
+func (i *inventoryManagerImpl) tryExecuteRebalance(ctx context.Context, rebalance *RebalanceData) (err error) {
+	ctx, span := i.handler.Tracer().Start(ctx, "tryExecuteRebalance", trace.WithAttributes(
+		attribute.Int("origin", rebalance.OriginMetadata.ChainID),
+		attribute.Int("dest", rebalance.OriginMetadata.ChainID),
```
ChiTimesChi, line 480: Should be `DestMetadata` here

services/rfq/relayer/inventory/rebalance.go
```diff
@@ -27,183 +27,185 @@ type RebalanceManager interface {
...
-		return nil, nil
-	}
+	return rebalances, nil
+}
 
-	// if the given chain is not the destination of the rebalance, no need to do anything
-	if destTokenData.ChainID != chainID {
-		if span != nil {
-			span.SetAttributes(attribute.Int("rebalance_dest", destTokenData.ChainID))
+// getRebalanceCandidates gets the best rebalance for each token and rebalance method supported by the config.
```
ChiTimesChi, line 53: I believe the documentation should state that it gets _all possible rebalances_ for each token rather than _the best rebalance_

services/rfq/relayer/inventory/rebalance.go
```diff
@@ -27,183 +27,185 @@ type RebalanceManager interface {
...
+func getRebalanceCandidates(ctx context.Context, cfg relconfig.Config, inv map[int]map[common.Address]*TokenMetadata) (rebalances map[string]map[relconfig.RebalanceMethod]*RebalanceData, err error) {
+	rebalances = map[string]map[relconfig.RebalanceMethod]*RebalanceData{}
+	for chainID, chainCfg := range cfg.Chains {
+		for tokenName, tokenCfg := range chainCfg.Tokens {
+			if len(tokenCfg.RebalanceMethods) == 0 {
+				continue
+			}
```
ChiTimesChi, line 60: Do we need this check? And if yes, should we check `methods` length below instead to avoid relying on `TokenConfig` structure?

services/rfq/relayer/inventory/rebalance.go
```diff
@@ -27,183 +27,185 @@ type RebalanceManager interface {
...
+		var candidateMetadata *TokenMetadata
+		for name, tokenCfg := range chainCfg.Tokens {
+			if name != tokenName {
+				continue
+			}
+
+			// check that the token supports given rebalance method
+			if supportsRebalanceMethod(cfg, chainID, tokenCfg.Address, method) {
+				validCandidate = true
+				candidateMetadata = inv[chainID][common.HexToAddress(tokenCfg.Address)]
```
ChiTimesChi, line 137: Can we just assign to `candidateChains[chainID]` here before breaking and avoid tracking `validCandidate` and `candidateMetadata`?

services/rfq/relayer/inventory/rebalance.go
```diff
@@ -27,183 +27,185 @@ type RebalanceManager interface {
...
-					}
-
-					// assign origin / dest metadata based on min / max balances
-					if originCandidate == nil || tokenData.Balance.Cmp(originCandidate.Balance) > 0 {
-						originCandidate = tokenData
-					}
-					if destCandidate == nil || tokenData.Balance.Cmp(destCandidate.Balance) < 0 {
-						destCandidate = tokenData
-					}
+// getCandidateChains gets the respective token metadata for each chain that supports the rebalance method.
```
ChiTimesChi, line 123: Change to "supports the rebalance method _for the given token._"?

services/rfq/relayer/inventory/rebalance.go
```diff
@@ -27,183 +27,185 @@ type RebalanceManager interface {
...
 }
 
-func isTokenCompatible(tokenData *TokenMetadata, method relconfig.RebalanceMethod, cfg relconfig.Config) bool {
-	// make sure that the token is compatible with our rebalance method
-	tokenMethods, tokenErr := cfg.GetRebalanceMethods(tokenData.ChainID, tokenData.Addr.Hex())
-	if tokenErr != nil {
-		logger.Errorf("could not get token rebalance method: %v", tokenErr)
-		return false
-	}
+func getBestRebalance(candidates []RebalanceData) (best *RebalanceData) {
```
ChiTimesChi, line 152: Function is self-explanatory, but a small description wouldn't hurt

services/rfq/relayer/inventory/rebalance.go
```diff
@@ -27,183 +27,185 @@ type RebalanceManager interface {
...
+		span.SetAttributes(attribute.String("max_token_balance", rebalance.OriginMetadata.Balance.String()))
+		span.SetAttributes(attribute.String("min_token_balance", rebalance.DestMetadata.Balance.String()))
 		span.SetAttributes(attribute.String("total_balance", totalBalance.String()))
 		span.SetAttributes(attribute.String("maintenance_thresh", maintenanceThresh.String()))
 	}
 
 	// no need to rebalance if we are not below maintenance threshold on destination
-	if destTokenData.Balance.Cmp(maintenanceThresh) > 0 {
+	if rebalance.DestMetadata.Balance.Cmp(maintenanceThresh) > 0 {
+		fmt.Println("returning nil")
```
ChiTimesChi, line 202: Remove the print here?

services/rfq/relayer/inventory/rebalance.go
```diff
@@ -27,183 +27,185 @@ type RebalanceManager interface {
...
 		return nil, nil
 	}
 
 	// calculate the amount to rebalance vs the initial threshold on destination
 	initialThresh, _ := new(big.Float).Mul(new(big.Float).SetInt(totalBalance), big.NewFloat(initialPct/100)).Int(nil)
-	amount = new(big.Int).Sub(originTokenData.Balance, initialThresh)
+	amount = new(big.Int).Sub(rebalance.OriginMetadata.Balance, initialThresh)
```
ChiTimesChi, line 208: @dwasse could you explain the logic here wrt initial value of the `amount`? Is the happy path here to rebalance exactly as much to get the destination chain balance back to the initial?

services/rfq/relayer/inventory/rebalance.go
```diff
@@ -252,3 +254,30 @@ func getRebalanceAmount(span trace.Span, cfg relconfig.Config, tokens map[int]ma
...
+			}
+			if supportsRebalanceMethod(cfg, tokenData.ChainID, tokenData.Addr.Hex(), method) {
+				totalBalance.Add(totalBalance, tokenData.Balance)
+			}
+		}
+	}
+	return totalBalance
+}
+
+func supportsRebalanceMethod(cfg relconfig.Config, chainID int, addr string, method relconfig.RebalanceMethod) bool {
```
ChiTimesChi, line 275: This feels generic enough to be located somewhere else, but I'm not entirely sure. Would trust the Go masters judgement here. In any case, could use a few words in the docs.

ChiTimesChi (changes requested):
First pass of the review. Didn't go through the test cases yet.

services/rfq/relayer/inventory/rebalance.go
dwasse: I reworked this function slightly and added some documentation:

```
// At a high level, there are three steps:
//  1. Evaluate the relevant maintenance / initial thresholds on origin and destination
//  2. Arrive at an initial rebalance amount, first targeting the initial threshold on origin,
//     and then clipping by the initial threshold on destination if necessary
//  3. Filter the rebalance amount by the configured min and max
```

In short, happy path is to bridge whatever amount takes the origin to initial threshold. If this number would take destination above initial threshold, we crop by that initial threshold as well

services/rfq/relayer/inventory/rebalance.go
dwasse: Perhaps a misleading function name- the intention is to get the best rebalance (taking origin / dest balances into account) for each token,method pair, to be filtered down across methods in `getRebalances()`. Note how `getRebalanceCandidates()` returns map[int]map[method]Rebalance, instead of the final value being a collection of rebalances

services/rfq/relayer/inventory/rebalance.go
dwasse: Only reason I do this check first is to avoid inserting an empty map value into `rebalances` in the case that there are no valid rebalances for a given token. Not critical but I think slightly cleaner to only deal with non-empty maps

services/rfq/relayer/inventory/rebalance.go
dwasse: Yeah, I mostly split this out to avoid dedup and make other functions a bit less verbose, I think it's probably fine to leave in this file as long as it's not used elsewhere

services/rfq/relayer/inventory/rebalance.go
```diff
@@ -27,218 +27,210 @@ type RebalanceManager interface {
...
 		return nil, nil
 	}
 
-	// calculate the amount to rebalance vs the initial threshold on destination
-	initialThresh, _ := new(big.Float).Mul(new(big.Float).SetInt(totalBalance), big.NewFloat(initialPct/100)).Int(nil)
-	amount = new(big.Int).Sub(originTokenData.Balance, initialThresh)
-
-	// no need to rebalance since amount would not be positive
+	// initially, set the rebalance amount such that it would take origin to the initial threshold
+	initialThreshOrigin, _ := new(big.Float).Mul(new(big.Float).SetInt(totalBalance), big.NewFloat(initialPctDest/100)).Int(nil)
```
ChiTimesChi, line 207: @dwasse based on https://github.com/synapsecns/sanguine/pull/3003#discussion_r1725744639 we should multiply by `initialPctOrigin` rather than `initialPctDest` here.

services/rfq/relayer/inventory/rebalance.go
ChiTimesChi: Oh, I see what you mean now, thanks. The name seems fine to me, just needed to reread the docs/impl one more time to get this.

PR #3045, opened 2024-08-21
refactor(relapi): deprecate `GetQuoteRequestStatusResponse` model

golangisfun123:
It was redundant and unclear, simply integrated the remaining fields into the existing `GetQuoteRequestResponse` model.


trajan0x (approved):

trajan0x (changes requested):
Can we delete the status routes as well?

https://github.com/synapsecns/sanguine/blob/b65003ca6c919cc5dc86d68d1cb2e34c08341c5a/services/rfq/relayer/relapi/server.go#L100-L101

If concern si brekaing backwards compatibility LGTM then delete in a future pr

dwasse (approved):

PR #3062, opened 2024-08-26
feat(rfq-relayer): wait for finality before proving

dwasse:
## Summary by CodeRabbit

- **New Features**
	- Introduced new configuration fields, `ProveConfirmations` and `FinalityConfirmations`, for enhanced control over confirmation processes in RFQ operations.
	- Added functions to retrieve `ProveConfirmations` and `FinalityConfirmations` values for specific chain configurations.
	- Implemented a new function to retrieve the relay block number for quote requests.

- **Improvements**
	- Enhanced the proof submission process to ensure sufficient confirmations are received before proceeding, improving reliability.
	- Added verification for finalized transactions in the processing logic, enhancing transaction handling robustness.
	- Incorporated a block number field in transaction models to improve tracking of blockchain-related events.


services/rfq/relayer/service/handlers.go
```diff
@@ -378,8 +378,28 @@ func (r *Relayer) handleRelayLog(parentCtx context.Context, req *fastbridge.Fast
...
+		attribute.Int("current_block_number", int(currentBlockNumber)),
+		attribute.Int("dest_block_number", int(receipt.BlockNumber.Int64())),
+		attribute.Int("prove_confirmations", int(proveConfirmations)),
+	)
+	if currentBlockNumber < request.BlockNumber+proveConfirmations {
+		span.AddEvent("not enough confirmations")
+		return nil
+	}
+
+	// relay has been finalized, it's time to go back to the origin chain and try to prove
```
ChiTimesChi, line 402: @dwasse we need to do a check similar to Guard's `isProveValid` here to handle the case where our relay tx was reorged.

In the event where we submitted a tx, witnessed a `BridgeRelayed` event, and later on this tx was reorged and ended up being reverted, we need a way for the Relayer to mark the tx as RelayRaceLost here (I'm not sure if we can safely assume that we will index the correct event with another relayer before that).

The current code will still get a valid receipt for `DestTxHash`, it's just the receipt will not have a corresponding `BridgeRelayed` event there.

PR #3084, opened 2024-08-30
[wip] feat(synapse-interface): confirm new price

bigboydiamonds:


packages/synapse-interface/components/StateManagedBridge/hooks/useConfirmNewBridgePrice.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  useEffect(() => {
+    const validQuotes =
+      bridgeQuote?.outputAmount && previousBridgeQuote?.outputAmount
+
+    const outputAmountDiffMoreThan1bps = validQuotes
+      ? calculateOutputRelativeDifference(
+          bridgeQuote,
+          quoteRef.current ?? previousBridgeQuote
+        ) > 0.0001
```
abtestingalpha, line 57: Worth putting this into a constant if we need to update

packages/synapse-interface/components/StateManagedBridge/BridgeTransactionButton.tsx
```diff
@@ -162,6 +178,13 @@ export const BridgeTransactionButton = ({
       onClick: () => switchChain({ chainId: fromChainId }),
       pendingLabel: 'Switching chains',
     }
+  } else if (hasQuoteOutputChanged && !hasUserConfirmedChange) {
+    buttonProperties = {
+      label: 'Confirm new price',
```
abtestingalpha, line 183: `bridgeModule` could change as well. Example: User first gets RFQ quote, and then inventory level changes. The quote is now through `CCTP`.

"Confirm new quote" (instead of price) covers both cases. Does the plumbing in the hook cover this as well?

packages/synapse-interface/utils/hooks/useStaleQuoteUpdater.ts
```diff
@@ -17,34 +16,42 @@ export const useStaleQuoteUpdater = (
...
-        if (eventListenerRef.current) {
-          document.removeEventListener('mousemove', eventListenerRef.current)
-        }
+    if (isQuoteValid && !isQuoteLoading && !isWalletPending) {
+      timeoutRef.current = setTimeout(() => {
+        eventListenerRef.current = null
+        setIsStale(true)
 
         const newEventListener = () => {
           refreshQuoteCallback()
```
abtestingalpha, line 35: One thing that comes to mind: If `refreshQuoteCallback` takes longer than `staleTimeout` to complete, you might end up with multiple queued refresh calls. Does the workflow take that into consideration?

packages/synapse-interface/components/StateManagedBridge/BridgeTransactionButton.tsx
bigboydiamonds: Good catch - confirming that the hook covers the scenario that `bridgeModule` changes while bridge selections remain the same as previous quote. 

The `useConfirmNewPrice()` hook will check for changes greater than 1bps given current and previous quotes share the same origin input amount, origin chainId, origin token, destination chainId, and destination token. 

Updated button text in f682d7337

packages/synapse-interface/utils/hooks/useStaleQuoteUpdater.ts
bigboydiamonds: At any given point in time, only one event listener will be attached in `eventListenerRef`. Multiple listeners will not queue up since the timeout and listener are cleaned up before the next event listener is set. 

This scenario is covered since the `setTimeout` function that attaches a new event listener will not fire off if a quote is currently loading. Once the `refreshQuoteCallback` completes (one that takes longer than `staleTimeout`), the new quote will trigger the useEffect to restart the workflow to check when to refresh again.

bigboydiamonds: Refactored in https://github.com/synapsecns/sanguine/commit/3b7cd9aee20f3e8e735bcdd76b63843ca97aaebd to improve legibility of workflow

PR #3098, opened 2024-09-04
Feat/eip191recov

parodime:
eip191 auth should accept recovery Id (aka "v") values of 0/1 or 27/28, as both methods are used inconsistently by various crypto libraries.

note that because the V value is not part of the actual encryption, we just normalize it before evaluating the signature

## Summary by CodeRabbit

- **New Features**
	- Enhanced handling of Ethereum signature recovery IDs for improved authentication robustness.
	- Added validation for signature normalization to accommodate various recovery ID formats.

- **Tests**
	- Introduced a new test to validate the normalization of recovery ID values during the EIP191 signature process, ensuring accurate authentication.

### DIFF
--- a/services/rfq/api/rest/server_test.go
+++ b/services/rfq/api/rest/server_test.go
@@ -67,6 +67,45 @@ func (c *ServerSuite) TestEIP191_SuccessfulSignature() {
 	c.Equal(http.StatusOK, resp.StatusCode)
 }
 
+// TestEIP191_SuccessfulSignature_vCodeNormalize tests the EIP191 signature process for successful authentication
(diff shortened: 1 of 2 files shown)


services/rfq/api/rest/server_test.go
```diff
@@ -67,6 +67,45 @@ func (c *ServerSuite) TestEIP191_SuccessfulSignature() {
...
+		return
+	}
+	defer func() {
+		err = resp.Body.Close()
+		c.Require().NoError(err)
+	}()
+
+	// Log the response body for debugging.
+	body, _ := io.ReadAll(resp.Body)
+	fmt.Println(string(body))
```
dwasse, line 103: We can remove this

dwasse (approved):

PR #3108, opened 2024-09-06
[SLT-131] +X-API-Version http header, +versions.json

parodime:
**Description**
Establishes a method of passively communicating our API version to all integrated parties.
Integrators will receive the version # as response header "X-API-Version" on each call.

This can be used by integrators to passively detect API version changes, at which point they can refer to our repository and versions.go file for additional information on the new version.

Note that versions.go not only includes our version #s and dates, but also allows us to add comments and alerts that will be visible to all.

We can also use this to publish alerts about deprecations or other important planned future changes, where perhaps no functionality is immediately changing.

**Additional context**
While Synapse may take additional steps to alert & advise on API changes, it will remain the responsibility of the integrators to set up their own detection & notifications of version changes as they use these endpoints. Subsequently, it will be their responsibility check our repository & versions.go file, to research & understand the changes being made and how it affects their integration, and to make any necessary code changes on their end.

**Metadata**
SLT-131

## Summary by CodeRabbit

- **New Features**
	- Introduced a new middleware that adds an `X-API-Version` header to API responses, enhancing version communication.
	- Added a mechanism to retrieve and log the current API version from a JSON file.
	- Created a `versions.json` file to manage and display versioning information for the API.
	- Established a structured representation of version information in a new `versions.go` file.

- **Tests**
	- Enhanced test coverage by validating the `X-API-Version` header in multiple API response scenarios.


services/rfq/api/versions.json
```diff
@@ -0,0 +1,11 @@
+{
```
dwasse, line 1: I would remove this file just for DRY purposes

services/rfq/api/rest/versions.go
```diff
@@ -0,0 +1,13 @@
+package rest
+
+var versionsJSON = `{
```
dwasse, line 3: This is a nit but IMO it would be cleaner to have a go struct defined here instead of parsing json

Something like:

```go
type version struct {
  version string
  date time.Time
  comment string
  alerts string
}

var versions = []version{
{
  Version: "1.0",
  Date: time.MustParse("2024-01-01"),
}
}
```

If we think we want to parse from json in the future can always add `json:` struct tags

dwasse:
Any chance we could verify that the version is set properly in one of the server tests?

services/rfq/api/versions.json
parodime: did not mean to leave that here - thanks!

services/rfq/api/rest/versions.go
parodime: agree - will change

trajan0x (approved):
lgtm

PR #3122, opened 2024-09-11
API method updates

Defi-Moses:
This adds the sdk functions getBridgeTxStatus and getSynapseTxId to the rest API, also adds a sub-method for finding the destination transaction information if a transaction is completed. 

docs are already updated. Examples could be significantly improved. 

## Summary by CodeRabbit

- **New Features**
	- Introduced three new API endpoints: 
		- `/getSynapseTxId` for retrieving a Synapse transaction ID.
		- `/getBridgeTxStatus` for checking the status of a bridge transaction.
		- `/getDestinationTx` for retrieving the destination transaction hash.
- **Error Handling**
	- Enhanced error handling and validation for all new endpoints to ensure robust responses.

### DIFF
--- a/packages/rest-api/src/app.ts
+++ b/packages/rest-api/src/app.ts
@@ -471,6 +471,114 @@ app.get('/bridgeTxInfo', async (req, res) => {
   }
 })
 
+// Get Synapse Transaction ID
+app.get('/getSynapseTxId', (req, res) => {
+  try {
+    const query = req.query
+    const originChainId = Number(query.originChainId)
+    const bridgeModuleName = String(query.bridgeModuleName)
+    const txHash = String(query.txHash)
+
+    if (!originChainId || !bridgeModuleName || !txHash) {
+      res.status(400).send({
+        message: 'Invalid request: Missing required parameters',
+      })
+      return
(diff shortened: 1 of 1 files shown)


packages/rest-api/src/app.ts
```diff
@@ -471,6 +471,116 @@ app.get('/bridgeTxInfo', async (req, res) => {
...
+    const bridgeModule = String(query.bridgeModule)
+    const synapseTxId = String(query.synapseTxId)
+
+    if (!destChainId || !bridgeModule || !synapseTxId) {
+      res.status(400).send({
+        message: 'Invalid request: Missing required parameters',
+      })
+      return
+    }
+
```
bigboydiamonds, line 523: If we wanted to prevent unnecessary SDK calls in the scenario where Client passes in an unsupported `destChainId`, we can do a check if a valid chainId is outputted from 

CHAINS_BY_ID[query.destChainId] 

from https://github.com/synapsecns/sanguine/blob/api/method-updates/packages/rest-api/src/constants/chains.ts/#L495

bigboydiamonds (approved):

PR #3133, opened 2024-09-16
REST API Improvements [SLT-179]

Defi-Moses:
Small fix to edit the arguments for the swaptxinfo to include the address so that approval estimation is correct

## Summary by CodeRabbit

- **New Features**
	- Enhanced validation for token addresses in bridge and swap transactions, ensuring only valid addresses are processed.
	- Improved error handling for unsupported tokens, providing clearer feedback to users.
	- Introduced utilities to validate token addresses and check token support on specific blockchains.
- **Bug Fixes**
	- Updated tests to use specific Ethereum token addresses for better accuracy in validation.
	- Enhanced error messaging in tests for invalid token addresses and parameters.

### DIFF
--- a/packages/rest-api/src/controllers/bridgeController.ts
+++ b/packages/rest-api/src/controllers/bridgeController.ts
@@ -1,6 +1,5 @@
 import { validationResult } from 'express-validator'
 import { parseUnits } from '@ethersproject/units'
-import { isAddress } from 'ethers/lib/utils'
 
 import { formatBNToString } from '../utils/formatBNToString'
 import { Synapse } from '../services/synapseService'
@@ -14,18 +13,8 @@ export const bridgeController = async (req, res) => {
   try {
     const { fromChain, toChain, amount, fromToken, toToken } = req.query
 
-    if (!isAddress(fromToken) || !isAddress(toToken)) {
-      return res.status(400).json({ error: 'Invalid token address' })
-    }
-
     const fromTokenInfo = tokenAddressToToken(fromChain.toString(), fromToken)
     const toTokenInfo = tokenAddressToToken(toChain.toString(), toToken)
-
-    if (!fromTokenInfo || !toTokenInfo) {
-      return res
-        .status(400)
-        .json({ error: 'Token not supported on specified chain' })
-    }
 
     const amountInWei = parseUnits(amount.toString(), fromTokenInfo.decimals)
 
--- a/packages/rest-api/src/controllers/bridgeController.ts
+++ b/packages/rest-api/src/controllers/bridgeController.ts
@@ -1,8 +1,10 @@
 import { validationResult } from 'express-validator'
 import { parseUnits } from '@ethersproject/units'
+import { isAddress } from 'ethers/lib/utils'
 
 import { formatBNToString } from '../utils/formatBNToString'
 import { Synapse } from '../services/synapseService'
+import { tokenAddressToToken } from '../utils/tokenAddressToToken'
 
 export const bridgeController = async (req, res) => {
   const errors = validationResult(req)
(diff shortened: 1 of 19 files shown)


abtestingalpha:
`address` is a new required field, so in that case need to 
* validate it in `routes/swapTxInfoRoute.ts` (currently only checking for other fields, need to validate new field)
* update test to `tests/swapTxInfoRoute.test.ts`
* run `yarn test` locally to make sure passing

Tests are currently failing bc no validation / missing address parameter

packages/rest-api/src/controllers/swapTxInfoController.ts
```diff
@@ -10,23 +12,34 @@ export const swapTxInfoController = async (req, res) => {
...
+      toToken,
       amountInWei
     )
 
     const txInfo = await Synapse.swap(
       Number(chain),
-      fromTokenInfo.address,
-      toTokenInfo.address,
+      address,
+      toToken,
```
ChiTimesChi, line 42: This has to be `fromToken`, the swap function takes the initial token/amount as a parameter (while the result token/min amountOut /deadline are provided in the query)

packages/rest-api/src/controllers/bridgeController.ts
```diff
@@ -1,26 +1,39 @@
...
   if (!errors.isEmpty()) {
     return res.status(400).json({ errors: errors.array() })
   }
   try {
-    const { fromChain, toChain, amount } = req.query
-    const fromTokenInfo = res.locals.tokenInfo.fromToken
-    const toTokenInfo = res.locals.tokenInfo.toToken
+    const { fromChain, toChain, amount, fromToken, toToken } = req.query
+
+    if (!isAddress(fromToken) || !isAddress(toToken)) {
```
abtestingalpha, line 17: if this is already checked in the route, it isn't needed here since the validations happen upstream in `bridgeRoute`

packages/rest-api/src/controllers/bridgeController.ts
```diff
@@ -1,26 +1,39 @@
...
+      return res.status(400).json({ error: 'Invalid token address' })
+    }
+
+    const fromTokenInfo = tokenAddressToToken(fromChain.toString(), fromToken)
+    const toTokenInfo = tokenAddressToToken(toChain.toString(), toToken)
+
+    if (!fromTokenInfo || !toTokenInfo) {
```
abtestingalpha, line 24: We can move this validation to the `bridgeRoute` (and in the other similar places this is used). Main reason to do this is our non-500 validations happen in the route so it's a clear contract for what the API expects.

abtestingalpha (changes requested):

packages/rest-api/src/controllers/swapTxInfoController.ts
```diff
@@ -10,23 +11,23 @@
   }
 
   try {
-    const { chain, amount } = req.query
-    const fromTokenInfo = res.locals.tokenInfo.fromToken
-    const toTokenInfo = res.locals.tokenInfo.toToken
+    const { chain, amount, address, fromToken, toToken } = req.query
```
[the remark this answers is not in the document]
Defi-Moses: This is included in swapTxInfoRoute.ts

packages/rest-api/src/routes/swapRoute.ts
```diff
@@ -17,8 +18,24 @@ router.get(
...
-    validateTokens('chain', 'toToken', 'toToken'),
+    check('fromToken')
+      .exists()
+      .withMessage('fromToken is required')
+      .custom((value) => isTokenAddress(value))
+      .withMessage('Invalid fromToken address')
+      .custom((value, { req }) =>
+        isTokenSupportedOnChain(value, req.query.chain as string)
+      )
+      .withMessage('Token not supported on specified chain'),
```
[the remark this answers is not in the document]
Defi-Moses: this is wrong, /swap doesnt accept an address

abtestingalpha (approved):

PR #3165, opened 2024-09-21
fix(api): formatted bridge fee amount [SLT-227]

bigboydiamonds:
## Summary by CodeRabbit

- **New Features**
	- Updated calculation method for the bridge fee, enhancing the accuracy of fee representation in the response payload. 

- **Bug Fixes**
	- Improved handling of decimal precision for bridge fees based on the latest token information.


Defi-Moses (approved):
Good catch, since the fee is taken in the intermediary token, that is the decimals that matter, not the final token.

Are there other places where this logic would be faulty?

PR #3170, opened 2024-09-23
store relayer on relay [SLT-182]

parodime:
## Description

`FastBridge` contract doesn’t save the Relayer address, who called `relay()` on the destination chain, which leaves the emitted logs as the only source of truth for this information. This makes the Guard/Relayer implementations a bit more fragile, as opposed to the situation where this information could’ve been read on-chain.

This change could be implemented without any additional gas costs by saving the relevant relayer information instead of a single bool flag in `bridgeRelays`.

- Save the `block.number, block.timestamp, relayer` used for completing the `relay()`
    - This would allow the Guards and Relayers to read this information and implement custom logic around block confirmations.
    - This could be packed in a single 32 bytes slot to replace `bridgeRelays` mapping.
- Expose backwards compatible `bridgeRelays(bytes32)` view.

NOTE: decided to not build a more integrator-friendly view that returns a struct of bridgeRelayDetails as opposed to default tuple -- this keeps it consistent with our other mappings.

## Summary by CodeRabbit

- **New Features**
	- Enhanced bridge transaction management with detailed relay information.
	- Introduced a new function to check the relay status of transactions.
	- Added a structure to store detailed relay transaction data, including block number, timestamp, and relayer address.

- **Bug Fixes**
	- Improved logic for verifying if a transaction has been relayed, ensuring accurate status checks.


packages/contracts-rfq/contracts/FastBridgeV2.sol
```diff
@@ -139,8 +139,13 @@ contract FastBridgeV2 is Admin, IFastBridgeV2 {
         if (block.timestamp > transaction.deadline) revert DeadlineExceeded();
 
         // mark bridge transaction as relayed
-        if (bridgeRelays[transactionId]) revert TransactionRelayed();
-        bridgeRelays[transactionId] = true;
+        if (bridgeRelayDetails[transactionId].blockNumber != 0) revert TransactionRelayed();
```
ChiTimesChi, line 142: To avoid repetition, let's make `bridgeRelays(bytes32)` a public view - will make it easier to maintain later on.

packages/contracts-rfq/contracts/interfaces/IFastBridgeV2.sol
```diff
@@ -20,4 +20,15 @@ interface IFastBridgeV2 is IFastBridge {
     /// @param request The encoded bridge transaction to claim on origin chain
     function claim(bytes memory request) external;
 
+    /// @notice Checks if a transaction has been relayed
```
ChiTimesChi, line 23: Needs `forge fmt` + in general we try to define the structs at the top of the contract, prior to the functions. In hindsight, both of these should've been flagged by CI workflows - I will make a ticket to update these with a reasonable settings (e.g. older `FastBridge` has a ton of linter warnings that might as well be ignored at this point)

packages/contracts-rfq/contracts/FastBridgeV2.sol
```diff
@@ -139,8 +139,13 @@ contract FastBridgeV2 is Admin, IFastBridgeV2 {
...
+        
+        bridgeRelayDetails[transactionId] = BridgeRelay({
+            blockNumber: uint48(block.number),
+            blockTimestamp: uint48(block.timestamp),
+            relayer: relayer
+        });
```
ChiTimesChi, line 148: This is a trivial change, but we still aim to get these covered in the tests. Both of these would be useful for testing:
- https://book.getfoundry.sh/cheatcodes/warp
- https://book.getfoundry.sh/cheatcodes/roll

ChiTimesChi: Also, I just realized that we don't have a unit test for `relayer == address(0)` case. Might as well add this one in destination-specific tests - and we could utilise this invariant in `bridgeRelays`

packages/contracts-rfq/contracts/FastBridgeV2.sol
```diff
@@ -173,6 +178,12 @@ contract FastBridgeV2 is Admin, IFastBridgeV2 {
             rebate
         );
     }
+    
+    /// @inheritdoc IFastBridgeV2
+    function bridgeRelays(bytes32 transactionId) external view returns (bool) {
+        // has this transactionId been relayed?
+        return bridgeRelayDetails[transactionId].blockNumber != 0;
```
ChiTimesChi, line 185: In my experience there was a very weird bug on one of the chains, where the block timestamp was set to zero. There is no real need in checking `block.number` and `block.timestamp` when saving in `relay()`, but just to be on the safer side of things - let's check `.relayer != address(0)` here? Once the test from the comment above is added, we could be 100% sure this works literally always.

ChiTimesChi (changes requested):

packages/contracts-rfq/contracts/FastBridgeV2.sol
parodime: doesnt seem like we're actually restricting relayer != address(0)

I'm adding that to the function as well - let me know if this is not what you had in mind

ChiTimesChi (approved):
LGTM with a minor nit and a merge conflict resolution

PR #3171, opened 2024-09-23
no empty sender recip [SLT-184]

parodime:
**Description**
Don’t allow empty sender/recipient addresses on bridge()

## Summary by CodeRabbit

- **New Features**
	- Enhanced input validation in the FastBridgeV2 contract to prevent transactions with zero addresses for key parameters.
  
- **Bug Fixes**
	- Updated tests to ensure they run correctly, validating behavior when sender or recipient addresses are zero.


packages/contracts-rfq/contracts/FastBridgeV2.sol
```diff
@@ -79,7 +79,12 @@ contract FastBridgeV2 is Admin, IFastBridgeV2 {
         // check bridge params
         if (params.dstChainId == block.chainid) revert ChainIncorrect();
         if (params.originAmount == 0 || params.destAmount == 0) revert AmountIncorrect();
-        if (params.originToken == address(0) || params.destToken == address(0)) revert ZeroAddress();
+        if (
+            params.originToken == address(0) || 
+            params.destToken == address(0) || 
+            params.sender == address(0) || 
+            params.to == address(0)
+        ) revert ZeroAddress();
```
ChiTimesChi, line 87: Does two independent checks make more sense to you here? If not, this needs a `// forgefmt: disable-next-item` before this statement to avoid being formatted in a very ugly way (try doing `forge fmt` and the result is much less readable than what it is now).

ChiTimesChi (approved):
LGTM with a minor styling nit

packages/contracts-rfq/contracts/FastBridgeV2.sol
parodime: makes sense to me for readability with forge fmt in mind, but not logical sense :)

i went ahead and made it 2 lines. Also confirmed via forge tests that it only increases gas by a couple units

ChiTimesChi: lgtm

PR #3173, opened 2024-09-23
status/proof mapping packing [SLT-186]

parodime:
## Description

A lot of functions are reading/writing from/to both `bridgeStatuses` and `bridgeProofs`, wasting more gas than necessary.

- Pack these into a new internal mapping to take a single storage slot
- Expose backwards-compatible views `bridgeStatuses(bytes32)` and `bridgeProofs(bytes32)`

## Summary by CodeRabbit

- **New Features**
  - Introduced a new structure for managing bridge transaction details, enhancing clarity and organization.
  - Added functions to retrieve transaction status and proof using transaction IDs.

- **Bug Fixes**
  - Updated references in tests to align with the new interface, ensuring consistent functionality.

- **Documentation**
  - Enhanced interface with new enumerations and structures for better tracking of bridge transaction statuses.

- **Tests**
  - Updated test files to incorporate the new interface and ensure proper functionality across the board.


trajan0x (approved):

packages/contracts-rfq/contracts/interfaces/IFastBridgeV2.sol
```diff
@@ -20,4 +20,33 @@ interface IFastBridgeV2 is IFastBridge {
...
+        REQUESTED,
+        RELAYER_PROVED,
+        RELAYER_CLAIMED,
+        REFUNDED
+    }
+
+    struct BridgeTxDetails {
+        BridgeStatus status;
+        BridgeProof proof;
+    }
```
ChiTimesChi, line 40: This will not have a desired effect of saving the gas on reads and writes, as the internal composition of the struct takes more than a single storage slot (256 bits). We don't export a gas report as of now, but in this case it's possible to generate using `forge t --mc FastBridgeV2SrcTest --nmt revert --gas-report`.
- Using `FastBridgeV2SrcTest` test contract gives us the overview for the source chain-related functions
- `--nmt revert` excludes test functions having a word "revert" in the name, which makes the report more useful (revert cases in general consume less gas, and thus need to be filtered)

ChiTimesChi: This are the results when run on master, prior to this PR
image

ChiTimesChi: And this are the results after this PR changes:
image

ChiTimesChi: To gain the performance in gas consumption, the `BridgeTxDetails` needs to fit within 256 bits:
- `BridgeStatus` enum is equivalent to `uint8`
- `address relayer` takes 160 bits
- which leaves up to 88 bits for the proof's timestamp

This is an opportunity to follow the suit of #3170 and also store the block number, when the prove happened. This could be useful for some of the Relayer implementations. This will require using up to 88 bits for both block number and timestamp, which is more than enough for practical purposes.

ChiTimesChi (changes requested):

parodime:
gas report after BridgeTxDetails refactor
2024-09-25 14_41_56-FastBridgeV2 sol - contracts-rfq  WSL_ Ubuntu  - SYNAPSE - Cursor

packages/contracts-rfq/contracts/FastBridgeV2.sol
```diff
@@ -183,31 +182,34 @@ contract FastBridgeV2 is Admin, IFastBridgeV2 {
...
         // update bridge tx status given proof provided
-        if (bridgeStatuses[transactionId] != BridgeStatus.REQUESTED) revert StatusIncorrect();
-        bridgeStatuses[transactionId] = BridgeStatus.RELAYER_PROVED;
-        bridgeProofs[transactionId] = BridgeProof({timestamp: uint96(block.timestamp), relayer: relayer}); // overflow ok
+        if (bridgeTxDetails[transactionId].status != BridgeStatus.REQUESTED) revert StatusIncorrect();
+        BridgeTxDetails storage _thisBridgeTxDetails = bridgeTxDetails[transactionId];
+        _thisBridgeTxDetails.status = BridgeStatus.RELAYER_PROVED;
+        _thisBridgeTxDetails.proofBlockTimestamp = uint40(block.timestamp);
+        _thisBridgeTxDetails.proofBlockNumber = uint48(block.number);
+        _thisBridgeTxDetails.proofRelayer = relayer;
```
ChiTimesChi, line 190: Why not just `bridgeTxDetails[transactionId].status = BridgeStatus.RELAYER_PROVED` and so on here?

packages/contracts-rfq/contracts/FastBridgeV2.sol
```diff
@@ -183,31 +182,34 @@ contract FastBridgeV2 is Admin, IFastBridgeV2 {
...
     /// @inheritdoc IFastBridge
     function canClaim(bytes32 transactionId, address relayer) external view returns (bool) {
-        if (bridgeStatuses[transactionId] != BridgeStatus.RELAYER_PROVED) revert StatusIncorrect();
-        BridgeProof memory proof = bridgeProofs[transactionId];
-        if (proof.relayer != relayer) revert SenderIncorrect();
-        return _timeSince(proof) > DISPUTE_PERIOD;
+        BridgeTxDetails storage _thisBridgeTxDetails = bridgeTxDetails[transactionId];
+        if (_thisBridgeTxDetails.status != BridgeStatus.RELAYER_PROVED) revert StatusIncorrect();
+        if (_thisBridgeTxDetails.proofRelayer != relayer) revert SenderIncorrect();
+        return _timeSince(_thisBridgeTxDetails.proofBlockTimestamp) > DISPUTE_PERIOD;
```
ChiTimesChi, line 212: Do we also need storage approach here? What if we do something like this instead?

`(BridgeStatus status, uint40 proofBlockTimestamp, , address proofRelayer) = bridgeTxDetails[transactionId]`

packages/contracts-rfq/contracts/FastBridgeV2.sol
```diff
@@ -220,21 +222,21 @@ contract FastBridgeV2 is Admin, IFastBridgeV2 {
         bytes32 transactionId = keccak256(request);
         BridgeTransaction memory transaction = getBridgeTransaction(request);
 
-        // update bridge tx status if able to claim origin collateral
-        if (bridgeStatuses[transactionId] != BridgeStatus.RELAYER_PROVED) revert StatusIncorrect();
+        BridgeTxDetails storage _thisBridgeTxDetails = bridgeTxDetails[transactionId];
```
ChiTimesChi, line 225: Not a huge fan of this tbh, makes the code more complex without a particularly good reason. Could either read the full proof in memory (as before), or unpack only the required variables (status, proofBlockTimestamp, proofRelayer) onto the stack as in the comment above.

packages/contracts-rfq/contracts/FastBridgeV2.sol
```diff
@@ -220,21 +222,21 @@ contract FastBridgeV2 is Admin, IFastBridgeV2 {
...
+            to = _thisBridgeTxDetails.proofRelayer;
+        } else if (_thisBridgeTxDetails.proofRelayer != msg.sender) {
             revert SenderIncorrect();
         }
 
-        if (_timeSince(proof) <= DISPUTE_PERIOD) revert DisputePeriodNotPassed();
+        if (_timeSince(_thisBridgeTxDetails.proofBlockTimestamp) <= DISPUTE_PERIOD) revert DisputePeriodNotPassed();
 
-        bridgeStatuses[transactionId] = BridgeStatus.RELAYER_CLAIMED;
+        _thisBridgeTxDetails.status = BridgeStatus.RELAYER_CLAIMED;
```
ChiTimesChi, line 239: My reasoning against the storage syntax is that it gets harder to track that this is a storage write rather than a memory one, which gets a bit confusing review-wise and could result in a bad change if the struct variable was ever made `memory` again.

ChiTimesChi (changes requested):
Not a fan of `storage` approach (see the comments) + needs merge conflicts to be resolved.

packages/contracts-rfq/contracts/FastBridgeV2.sol
parodime: my rationale for the storage reference was gas savings (and consistency, for the external views) -- but after using forge to test, the savings seemed minimal. switched to fully-qualified approach everywhere

packages/contracts-rfq/contracts/FastBridgeV2.sol
```diff
@@ -44,6 +34,16 @@ contract FastBridgeV2 is Admin, IFastBridgeV2, IFastBridgeV2Errors {
...
     uint256 public immutable deployBlock;
 
+    function bridgeStatuses(bytes32 transactionId) public view returns (BridgeStatus status) {
+        return bridgeTxDetails[transactionId].status;
+    }
+
+    function bridgeProofs(bytes32 transactionId) public view returns (uint96 timestamp, address relayer) {
+        uint96 proofBlockTimestamp = bridgeTxDetails[transactionId].proofBlockTimestamp;
+        address proofRelayer = bridgeTxDetails[transactionId].proofRelayer;
+        return (proofBlockTimestamp, proofRelayer);
```
ChiTimesChi, line 44: When you're using named return variables, you could just assign values to them and skip the explicit return statement - this would be a bit cleaner:
```suggestion
        timestamp = bridgeTxDetails[transactionId].proofBlockTimestamp;
        relayer = bridgeTxDetails[transactionId].proofRelayer;
```

packages/contracts-rfq/contracts/FastBridgeV2.sol
```diff
@@ -44,6 +34,16 @@ contract FastBridgeV2 is Admin, IFastBridgeV2, IFastBridgeV2Errors {
     // @dev the block the contract was deployed at
     uint256 public immutable deployBlock;
 
+    function bridgeStatuses(bytes32 transactionId) public view returns (BridgeStatus status) {
+        return bridgeTxDetails[transactionId].status;
+    }
+
+    function bridgeProofs(bytes32 transactionId) public view returns (uint96 timestamp, address relayer) {
```
ChiTimesChi, line 41: Let's leave the constructor as a first defined function, moving these all the way below would be canonical.

packages/contracts-rfq/test/FastBridgeV2.Src.t.sol
```diff
@@ -1,7 +1,21 @@
...
+    ChainIncorrect,
+    DisputePeriodNotPassed,
+    DisputePeriodPassed,
+    DeadlineNotExceeded,
+    DeadlineTooShort,
+    MsgValueIncorrect,
+    SenderIncorrect,
+    StatusIncorrect,
+    ZeroAddress
+} from "../contracts/libs/Errors.sol";
```
ChiTimesChi, line 15: There's a somewhat big inheritance chain in the test contracts, but eventually it gets to the `FastBridgeV2Test`, which inherits from `IFastBridgeV2Errors`. This small hack allows to have access to the errors (and their selectors) without having to import them from the library.

packages/contracts-rfq/contracts/interfaces/IFastBridgeV2.sol
```diff
@@ -4,6 +4,21 @@ pragma solidity ^0.8.20;
...
+        REQUESTED,
+        RELAYER_PROVED,
+        RELAYER_CLAIMED,
+        REFUNDED
+    }
+
+    struct BridgeTxDetails {
+        BridgeStatus status;
+        uint40 proofBlockTimestamp;
+        uint48 proofBlockNumber;
```
ChiTimesChi, line 18: This is purely hypothetical, but wouldn't using `uint48` for timestamps and `uint40` for block numbers make more sense here? The EVM chains have under 1 block/second, so `block.number < block.timestamp`.

Again, this is purely hypothetical, as overflowing these values is sooo far in the future.

ChiTimesChi (approved):
Solid work! I'm outlining some final styling nits, functionality-wise LGTM. Feel free to merge once the comments are resolved.

packages/contracts-rfq/contracts/interfaces/IFastBridgeV2.sol
parodime: my thought was that seconds timestamp is the more behaviorally predictable of the two and uint40 covers it sufficiently - so whatever is remaining can just go to Block #. AFAIK an L2 can technically decide to do whatever they want with their block #s, so overflow seemed more possible there (although in either case, the possibility is so remote to be basically nothing)

packages/contracts-rfq/contracts/FastBridgeV2.sol
parodime: NP. if cool w/ you, i'll rearrange this and most of the other components of FastBridgeV2 in a separate PR after this merges that focuses on eliminating all solhint warnings

packages/contracts-rfq/contracts/FastBridgeV2.sol
parodime: much cleaner - thanks

packages/contracts-rfq/contracts/FastBridgeV2.sol
ChiTimesChi: Sounds good

packages/contracts-rfq/contracts/interfaces/IFastBridgeV2.sol
ChiTimesChi: That's a very good point

PR #3214, opened 2024-10-02
FbV2 sender nonces [SLT-183]

parodime:
Description
Replace chain-scoped nonce with one that increments per-sender

Additional context
Reduces risk of reorgs resulting in a different nonce

## Summary by CodeRabbit

- **New Features**
	- Enhanced nonce management for improved security, allowing unique tracking per sender.
	- Deprecated the old nonce mechanism, introducing a new public function that always returns zero.

- **Bug Fixes**
	- Added assertions to ensure correct nonce incrementing for users during bridge operations.

- **Documentation**
	- Updated comments to reflect changes in nonce management and deprecation notices.

### DIFF
--- a/packages/contracts-rfq/.gitignore
+++ b/packages/contracts-rfq/.gitignore
@@ -1,3 +1,5 @@
 flattened
 .deployments
 broadcast
+
+lcov.info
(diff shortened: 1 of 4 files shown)


ChiTimesChi (approved):
LGTM

PR #3216, opened 2024-10-02
feat(rest-api): Adds winston logger

abtestingalpha:
**Description**
Adds `winston` for logging. Logs will be streamed as json to Railway.


Defi-Moses (approved):
Gave an approval/tested but couple of qs

1. Do we not want to log the request along with whatever is returned ? Would be a lot more information logged but may be easier to read/see errors

2. We dont log errors caught by  /routes whats the logic here ? Not sure how helpful this would be but would be interesting to see around what type of requests people make that are failing. 

Just trying to think a bit about what we want to accomplish here, how to make this useful

PR #3232, opened 2024-10-04
proof/dispute active & other tweaks/suggs [SLT-314]

parodime:
added concept of "active only" to `qProofs` and `qDisputes` -- so we can supply a parameter that will either include or exclude them based on what we need for each particular endpoint.

"Active" meaning whether the given `Proof` or `Dispute` has been supplanted by something more recent.
EG: if I submit a `Proof` and it is followed by a `Dispute`, then the `Proof` would be inactive and the `Dispute` would be active

If i then submit a new `Proof`, then that `Proof` becomes Active and the `Dispute` becomes inactive

This can repeat ad nauseum

Also lots of documentation & other tweaks

## Summary by CodeRabbit

- **New Features**
	- Enhanced API documentation with direct links to Swagger for all API calls.
	- Added environment variable details for development and database connections.

- **Improvements**
	- Updated API queries to focus on active entries for proofs and disputes, improving data relevance.
	- Renamed fields in the GraphQL schema for better clarity and consistency.
	- Introduced a new Google Analytics component for improved tracking.

- **Bug Fixes**
	- Corrected response messages to reflect active status in various API calls.

- **Documentation**
	- Improved clarity and organization of the RFQ Indexer API README.
	- Updated changelog entries for version tracking.


packages/rfq-indexer/api/src/queries/proofsQueries.ts
```diff
@@ -1,15 +1,25 @@
...
       'BridgeProofProvidedEvents.blockNumber as blockNumber_proof',
       'BridgeProofProvidedEvents.blockTimestamp as blockTimestamp_proof',
       'BridgeProofProvidedEvents.transactionHash as transactionHash_proof',
-
       'BridgeProofProvidedEvents.relayer as relayer_proof',
-    ])
-}
+    ]);
+
+  if (activeOnly) {
```
Defi-Moses, line 20: If there is a proof, and ensuing dispute, and then a new proof, woudl the new proof appear here? given that we are filtering for transactionId of the relayed transaction, wouldnt the new (valid) proof not appear?

Could be wrong here

Defi-Moses: Nevermind, the leftjoin solves for this

packages/rfq-indexer/api/src/graphql/types/events.graphql
```diff
@@ -86,6 +86,6 @@ scalar BigInt
     blockNumber: BigInt!
     blockTimestamp: Int!
     transactionHash: String!
-    originChainId: Int!
-    originChain: String!
+    chainId: Int!
```
Defi-Moses, line 89: Any reasoning behind this change? The only reason I did this was for standardization with all of our other queries (makes it neater to map fields). 

Not a big deal

Defi-Moses (commented):
Left a few thoughts but looks good to me. Feel free to merge whenever

PR #3246, opened 2024-10-07
feat(contracts-rfq): arbitrary call with value [SLT-233] [SLT-318]

ChiTimesChi:
**Description**
This PR finishes the work started in #3215 by adding the functionality to request a native value to be supplied on the destination chain, when relaying the request.

The requested "call value" will be attached to the recipient hook function call (if `callParams` is a non-empty payload), or transferred as a separate native transfer after the tokens are sent.

## Summary by CodeRabbit

- **New Features**
	- Enhanced bridging functionality with new parameters for handling call values and call parameters.
	- Introduced new error handling for unsupported native token call values.
	- Improved clarity in event emissions with named parameters.

- **Bug Fixes**
	- Improved logic for relaying transactions to ensure correct handling of call values and reverts.

- **Tests**
	- Expanded testing framework with additional scenarios for relaying tokens and ETH, including edge cases for call values and parameters.
	- Enhanced tests for various recipient behaviors regarding call values.


trajan0x (changes requested):
Have one more thought here. Would it make sense to add `version` as the first field to `BridgeTransactionV2` struct?

packages/contracts-rfq/contracts/FastBridgeV2.sol
```diff
@@ -371,7 +355,7 @@ contract FastBridgeV2 is Admin, IFastBridgeV2, IFastBridgeV2Errors {
         bytes memory hookData =
             abi.encodeCall(IFastBridgeRecipient.fastBridgeTransferReceived, (token, amount, callParams));
         // This will bubble any revert messages from the hook function
-        bytes memory returnData = Address.functionCallWithValue({target: recipient, data: hookData, value: msgValue});
+        bytes memory returnData = Address.functionCallWithValue({target: recipient, data: hookData, value: msg.value});
         // Explicit revert if no return data at all
         if (returnData.length == 0) revert RecipientNoReturnValue();
```
parodime, line 360: this would also confirm that the recipient must be a smart contract, not an EOA - correct?

packages/contracts-rfq/contracts/FastBridgeV2.sol
```diff
@@ -335,33 +324,28 @@ contract FastBridgeV2 is Admin, IFastBridgeV2, IFastBridgeV2Errors {
...
+        if (token == UniversalTokenLib.ETH_ADDRESS) {
+            // For ETH we just need to check that the supplied msg.value is correct
+            if (amount != msg.value) revert MsgValueIncorrect();
+            amountPulled = msg.value;
+        } else {
             token.assertIsContract();
             // Record token balance before transfer
-            amountPulled = IERC20(token).balanceOf(recipient);
+            amountPulled = IERC20(token).balanceOf(address(this));
             // Token needs to be pulled only if msg.value is zero
```
parodime, line 337: is this comment right? seems like msg.value can be !=0 and token still pulled if its an ERC20 bridge w/ callValue

parodime:
> @trajan0x this makes sense to me in terms of making the struct future-proof. I would prefer to have this done in a separate PR, however, as it's not just the extra field, but also we wouldn't want to accept the "future versions" of the struct in the state-changing functions.

wouldnt incompatible payloads result in reverts anyway? what else is gained by an explicit payload version param that I'm not thinking of?

ChiTimesChi:
> > @trajan0x this makes sense to me in terms of making the struct future-proof. I would prefer to have this done in a separate PR, however, as it's not just the extra field, but also we wouldn't want to accept the "future versions" of the struct in the state-changing functions.
> 
> wouldnt incompatible payloads result in reverts anyway? what else is gained by an explicit payload version param that I'm not thinking of?

This depends on the V3 payload internal structure. For instance, the initial V2 struct (without `bytes callParams`) was still decodable through `getBridgeTransaction()`. This happened because Solidity's `abi.decode` simply discards any extra data on top of the "expected length". For example:

```solidity
  bytes memory request = abi.encode(1, 2, 3);
  (uint256 a, uint256 b) = abi.decode(request, (uint256, uint256));
  // a == 1
  // b == 2
  // extra data with "3" is discarded
```

packages/contracts-rfq/contracts/interfaces/IFastBridgeV2.sol
```diff
@@ -30,14 +30,17 @@ interface IFastBridgeV2 is IFastBridge {
     /// for backwards compatibility.
     /// Note: quoteRelayer and quoteExclusivitySeconds are either both zero (indicating no exclusivity)
     /// or both non-zero (indicating exclusivity for the given period).
+    /// Note: callValue > 0 can NOT be used with destToken = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE (ETH_ADDRESS)
```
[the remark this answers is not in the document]
ChiTimesChi: This is done to reduce complexity on the Relayer side. Any request for `destAmount` worth of ETH with `callValue` on top of that could be reformulated as `destAmount + callValue` worth of ETH on the sender side.

For relayers this means that msg.value for the relay call is always:
- `destAmount` for ETH
- `callValue` for ERC20s

packages/contracts-rfq/contracts/FastBridgeV2.sol
ChiTimesChi: Good catch - seems that this comment (with no value check) is present in `FastBridge` as well. More revert cases for incorrectly invoked `bridge()` by the users to cover that would be nice, will make a separate ticket for that.

As for the `_pullToken`, it probably needs a rename to signal that it's only used in `bridge()` now.

packages/contracts-rfq/contracts/FastBridgeV2.sol
ChiTimesChi: Correct! This is handled by the OZ library here & here:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v5.0/contracts/utils/Address.sol#L83-L89

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v5.0/contracts/utils/Address.sol#L114-L129

packages/contracts-rfq/contracts/FastBridgeV2.sol
ChiTimesChi: @parodime the comment was fixed to be enforced in #3251, while the function itself got refactored in 189931c968e016c2f9d1f431ad8aebace124f6d7. Does this look cleaner now?

ChiTimesChi:
@coderabbitai summary

ChiTimesChi:
@coderabbitai full review

ChiTimesChi:
Requesting a follow-up review from @parodime and @trajan0x.

The inclusion of the version in the request payload is now tracked as SLT-328 and will be done in a separate PR.

trajan0x (approved):

parodime (approved):
LGTM

PR #3277, opened 2024-10-12
RFQ Indexer: adding search by origin tx

Defi-Moses:
Adding search by origin txn hash and not just transactionID. 

## Summary by CodeRabbit

- **New Features**
	- Enhanced transaction retrieval to support both transaction ID and transaction hash.
- **Documentation**
	- Updated OpenAPI documentation to clarify the dual functionality of the `/transaction-id/{transactionId}` endpoint.

### DIFF
--- a/packages/rfq-indexer/api/src/controllers/transactionIdController.ts
+++ b/packages/rfq-indexer/api/src/controllers/transactionIdController.ts
@@ -10,7 +10,12 @@ export const getTransactionById = async (req: Request, res: Response) => {
   try {
     const query = db
       .with('deposits', () =>
-        qDeposits().where('transactionId', '=', transactionId as string)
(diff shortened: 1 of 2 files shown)


parodime (approved):
looks great - i like the approach where its the same single input param & just checks both. I think this could even be expanded to just search for that hash *anywhere* -- transactionId, deposit, relay, proof, claim, dispute.

Another edge case catch is that unlike transactionId, a single transaction hash can technically repeat on multiple bridges. (EG: a single transaction can have any number of bridge deposits --- or relays, proofs, etc)

Not sure what current expected behavior of this endpoint would be in that case, but its probably best if it returns all of the matches, even if they are multiple bridges.

Its such an edge case tho right now -- I think this is good as-is & we can always do more later!

PR #3299, opened 2024-10-16
feat(rfq-relayer): fee pricer considers v2 CallValue and CallParams [SLT-320]

dwasse:
## Summary by CodeRabbit

- **New Features**
	- Enhanced fee calculation logic to include transaction-specific details for more accurate estimations.
	- Integration of transaction data into quote generation and fee calculations.
	- Improved testing framework for bridging functionality with enhanced concurrency handling.
	- Added support for additional parameters in bridge transactions to enhance functionality.
	- Introduced a new method for generating active RFQ messages.
	- New contract deployment mechanism for the bridge transaction contract.
	- New method for encoding and decoding bridge transactions.

- **Bug Fixes**
	- Improved error reporting and handling in HTTP response processing.

- **Documentation**
	- Improved comments to clarify the purpose of new parameters and methods.


services/rfq/relayer/pricer/fee_pricer.go
```diff
@@ -147,11 +157,95 @@ func (f *feePricer) GetDestinationFee(parentCtx context.Context, _, destination
...
+		span.SetAttributes(attribute.String("call_fee", callFee.String()))
+	}
+
+	if tx.CallValue != nil && tx.CallValue.Sign() > 0 {
+		callValueFloat := new(big.Float).SetInt(tx.CallValue)
+		valueDenom, err := f.getDenomFee(ctx, destination, destination, denomToken, callValueFloat)
+		if err != nil {
+			return nil, err
+		}
+		valueScaled, err := f.getFeeWithMultiplier(ctx, destination, isQuote, valueDenom)
```
ChiTimesChi, line 196: @dwasse I think this will be using `relay_fixed_fee_multiplier` when checking whether to relay a submitted user transaction. This value is under `1.0`, and therefore opens up a griefing vector (e.g. ask for  `CallValue` worth $1, but only supply $0.8 in fees)

ChiTimesChi: I think one of these two approaches should work here:
1. Always use `quote_fixed_fee_miltiplier` for determining how much is the requested `CallValue` worth
2. Introduce a separate multiplier parameter for that

services/rfq/relayer/pricer/fee_pricer.go
```diff
@@ -147,11 +157,95 @@ func (f *feePricer) GetDestinationFee(parentCtx context.Context, _, destination
...
+		return 0, fmt.Errorf("could not encode function call: %w", err)
+	}
+
+	callMsg := ethereum.CallMsg{
+		From:  f.relayerAddress,
+		To:    &tx.DestRecipient,
+		Value: tx.CallValue,
+		Data:  encodedData,
+	}
+	gasEstimate, err = client.EstimateGas(ctx, callMsg)
```
ChiTimesChi, line 240: We should be estimating the gas cost for the whole `relay()` call here instead.

Estimating gas for the recipient call should _probably_ work for the native gas token (assuming the recipient doesn't implement any caller checks, e.g. only allows the FastBridge to call the hook function), but will definitely fail for ERC20s. These are transferred separately by the `FastBridge` contract (as the ERC20 standard doesn't have built-in "transfer and call"), so just calling the hook function will result in a revert (as the recipient doesn't yet have the tokens transferred to them), leading to infinite gas estimation (or an error - not sure about the exact behaviour there).

ChiTimesChi (changes requested):
Commented above for the changes

services/rfq/relayer/pricer/fee_pricer.go
```diff
@@ -127,14 +137,28 @@ func (f *feePricer) GetDestinationFee(parentCtx context.Context, _, destination
...
 
-	// Calculate the destination fee
-	gasEstimate, err := f.config.GetDestGasEstimate(int(destination))
-	if err != nil {
-		return nil, fmt.Errorf("could not get dest gas estimate: %w", err)
+	fee := big.NewInt(0)
+
+	// Calculate the static L2 fee if it won't be incorporated by directly estimating the relay() call
+	// in addZapFees().
+	if quoteRequest == nil || quoteRequest.Transaction.CallParams == nil {
```
ChiTimesChi, line 144: Dont we need to check the `CallParams` length for being zero here? This should be present in every V2 quote request, the difference is whether it's empty (no Zap call requested) vs non-empty (providing data for the Zap call).

services/rfq/relayer/pricer/fee_pricer.go
```diff
@@ -147,11 +171,92 @@ func (f *feePricer) GetDestinationFee(parentCtx context.Context, _, destination
...
+		return 0, fmt.Errorf("could not get RFQ address: %w", err)
+	}
+
+	callMsg := ethereum.CallMsg{
+		From:  f.relayerAddress,
+		To:    &rfqAddr,
+		Value: quoteRequest.Transaction.CallValue,
+		Data:  encodedData,
+	}
+	gasEstimate, err = client.EstimateGas(ctx, callMsg)
```
ChiTimesChi, line 251: What's the expected of `EstiamateGas` here if the transaction reverts to matter what? For instance another relayer might've relayed this sooner than us, causing the execution simulation to fail. I'm not sure if this throws an error or returns an absurdly big return value here.

ChiTimesChi (commented):
Highlighting a few new questions from the comments

services/rfq/relayer/pricer/fee_pricer.go
dwasse: In my tests it returns an error. I think that's desired here? We could also have a fallback mechanism, though I'm not sure what that would look like.

services/rfq/relayer/pricer/fee_pricer.go
dwasse: Good call, I'm not sure how the ABI impl handles this but better safe than sorry

services/rfq/relayer/pricer/fee_pricer.go
```diff
@@ -147,11 +171,92 @@ func (f *feePricer) GetDestinationFee(parentCtx context.Context, _, destination
...
+
+	if fastBridgeV2ABI == nil {
+		parsedABI, err := abi.JSON(strings.NewReader(fastbridgev2.IFastBridgeV2MetaData.ABI))
+		if err != nil {
+			return 0, fmt.Errorf("could not parse ABI: %w", err)
+		}
+		fastBridgeV2ABI = &parsedABI
+	}
+
+	encodedData, err := fastBridgeV2ABI.Pack(methodName, quoteRequest.RawRequest, f.relayerAddress)
```
ChiTimesChi, line 235: Seeing an issue here. I don't think `quoteRequest.RawRequest` is getting set at all when the active quote is generated:

https://github.com/synapsecns/sanguine/blob/f93b3b375df843f372f7682aff9edb3de9da190d/services/rfq/relayer/quoter/quoter.go#L374-L394

Therefore we don't correctly estimate the gas here. Constructing the full request is somewhat tricky here (e.g. we need to set the destination amount in advance), happy to discuss the details on a call.

services/rfq/relayer/pricer/fee_pricer_test.go
```diff
@@ -207,21 +211,55 @@ func (s *PricerSuite) TestGetTotalFee() {
...
+	// Calculate the total fee with v2 call value and call params.
+	quoteRequest = &reldb.QuoteRequest{
+		RawRequest: []byte{},
+		Transaction: fastbridgev2.IFastBridgeV2BridgeTransactionV2{
+			ZapNative:  big.NewInt(1 * 1e18),
+			ZapData:    []byte{1},
+			DestAmount: big.NewInt(1 * 1e18),
+		},
+	}
+	clientDestination.On(testsuite.GetFunctionName(clientDestination.EstimateGas), mock.Anything, mock.Anything).Once().Return(uint64(1_000_000), nil)
```
ChiTimesChi, line 254: Likewise, using empty `RawRequest` during the on-chain call will result in a revert, so we should be using a more accurate mocking strategy (if any). My bet would be once the pricer correctly constructs the `RawRequest`, the gas estimation will just work without any mocking required.

ChiTimesChi (changes requested):
Need to complete the implementation for generating active quotes with Zaps

services/rfq/relayer/pricer/fee_pricer.go
dwasse: So we actually do supply the `RawRequest` when doing the `EstimateGas()` call here:

https://github.com/synapsecns/sanguine/blob/feat/arb-call-fee-pricer/services/rfq/relayer/pricer/fee_pricer.go#L235

The `QuoteInput.QuoteRequest` is really only used to communicate the extra data around the quoter functions. I considered introducing a new struct for this, maybe simpler to just include these two fields individually within the `QuoteInput` struct?

services/rfq/relayer/pricer/fee_pricer_test.go
dwasse: This is just a unit test with no anvil backend for on-chain interactions- IMO we are fine to mock here since the goal of these tests is just to make sure that the calls are made properly- anything relying on the actual chain interactions should be an e2e test

services/rfq/relayer/pricer/fee_pricer.go
ChiTimesChi: We supply it, but we never actually set the value for it, so it's always an empty bytes array. Let's figure out the best way to generate it, as the current contracts don't expose the encoding function.

ChiTimesChi (changes requested):
Final round of review. Also there are some merge conflicts that need to be fixed.

PR #3303, opened 2024-10-17
storage refs - readability & gas impv [SLT-353]

parodime:
- Improve references to bridgeTxDetails[transactionId] throughout the contract
- Theory is that storage refs of the main struct will be readable & efficient, so will try this approach first (borrowing the ‘$’ labeling convention from Cortex)
- If storage ref approach does not work out, use dedicated declared vars

## Summary by CodeRabbit

- **Chores**
	- Improved code readability and maintainability in the FastBridgeV2 contract by reducing redundancy in accessing transaction details.
	- Updated comments for better clarity without changing core functionality.

### DIFF
--- a/packages/contracts-rfq/contracts/FastBridgeV2.sol
+++ b/packages/contracts-rfq/contracts/FastBridgeV2.sol
@@ -88,7 +88,7 @@ contract FastBridgeV2 is Admin, IFastBridgeV2, IFastBridgeV2Errors {
             revert DisputePeriodPassed();
         }
 
-        // @dev relayer gets slashed effectively if dest relay has gone thru
+        // Note: these are storage writes
         $.status = BridgeStatus.REQUESTED;
         $.proofRelayer = address(0);
         $.proofBlockTimestamp = 0;
@@ -115,7 +115,7 @@ contract FastBridgeV2 is Admin, IFastBridgeV2, IFastBridgeV2Errors {
             if (block.timestamp <= transaction.deadline + REFUND_DELAY) revert DeadlineNotExceeded();
         }
 
-        // if all checks passed, set to REFUNDED status
+        // Note: this is a storage write
         $.status = BridgeStatus.REFUNDED;
 
         // transfer origin collateral back to original sender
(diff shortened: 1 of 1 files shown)


packages/contracts-rfq/contracts/FastBridgeV2.sol
```diff
@@ -77,18 +77,20 @@ contract FastBridgeV2 is Admin, IFastBridgeV2, IFastBridgeV2Errors {
...
-        if (_timeSince(bridgeTxDetails[transactionId].proofBlockTimestamp) > DISPUTE_PERIOD) {
+        BridgeTxDetails storage $ = bridgeTxDetails[transactionId];
+
+        if ($.status != BridgeStatus.RELAYER_PROVED) revert StatusIncorrect();
+        if (_timeSince($.proofBlockTimestamp) > DISPUTE_PERIOD) {
             revert DisputePeriodPassed();
         }
 
-        address disputedRelayer = bridgeTxDetails[transactionId].proofRelayer;
+        address disputedRelayer = $.proofRelayer;
```
ChiTimesChi, line 87: One extra change that allows the compiler optimizer to do its job. If you do multiple storage reads in a row (without any instructions in-between), I believe it is able to actually read it just once, if all the variables reside in the same storage slot (and they do, since it's a packed struct to fit within a single storage slot).

You can check this by running a gas bench against the optimized suggestion, and then moving the `disputedRelayer` assignment back after the checks - you should notice a slight gas increase.

```suggestion
        BridgeStatus status = $.status;
        uint40 proofBlockTimestamp = $.proofBlockTimestamp;
        address disputedRelayer = $.proofRelayer;
        if (status != BridgeStatus.RELAYER_PROVED) revert StatusIncorrect();
        if (_timeSince(proofBlockTimestamp) > DISPUTE_PERIOD) {
            revert DisputePeriodPassed();
        }
```

packages/contracts-rfq/contracts/FastBridgeV2.sol
```diff
@@ -288,21 +296,23 @@ contract FastBridgeV2 is Admin, IFastBridgeV2, IFastBridgeV2Errors {
         bytes32 transactionId = keccak256(request);
         BridgeTransactionV2 memory transaction = getBridgeTransactionV2(request);
 
+        BridgeTxDetails storage $ = bridgeTxDetails[transactionId];
```
ChiTimesChi, line 299: Similarly here, if we were to do all the `$` reads at this point into the stack variables, we would see some extra gas savings on top.

ChiTimesChi (approved):

PR #3336, opened 2024-10-25
exclusivityEndtime default 0 [SLT-391]

parodime:
Previously if bridge was not intended to be exclusive it would still end up with exclusivityEndtime = block.timestamp as a default from excl offset being 0.

This is generally fine & efficient, except it adds reorg risk to even non-exclusive relays whereby the bridge params are dependent upon block.timestamp and will change if reorged into a different block.

This approach will instead pay slightly more gas to default to a static exclusivityEndtime value (0) if no exclusivity is intended - inferred by quoteRelayer being address 0

## Summary by CodeRabbit

- **New Features**
	- Enhanced handling of bridge transaction exclusivity with updated conditions for `exclusivityEndTime`.

- **Bug Fixes**
	- Improved validation logic for exclusivity parameters during bridge transactions.

- **Tests**
	- Refined test function names for clarity regarding exclusivity checks.
	- Adjusted handling of exclusivity end times in gas benchmarking tests.

### DIFF
--- a/packages/contracts-rfq/contracts/FastBridgeV2.sol
+++ b/packages/contracts-rfq/contracts/FastBridgeV2.sol
@@ -177,7 +177,7 @@ contract FastBridgeV2 is Admin, IFastBridgeV2, IFastBridgeV2Errors {
 
     /// @inheritdoc IFastBridgeV2
     function bridge(BridgeParams memory params, BridgeParamsV2 memory paramsV2) public payable {
-        int256 exclusivityEndTime = paramsV2.quoteExclusivitySeconds != 0
+        int256 exclusivityEndTime = paramsV2.quoteRelayer != address(0)
             ? int256(block.timestamp) + paramsV2.quoteExclusivitySeconds
             : int256(0);
         _validateBridgeParams(params, paramsV2, exclusivityEndTime);
--- a/packages/contracts-rfq/contracts/FastBridgeV2.sol
+++ b/packages/contracts-rfq/contracts/FastBridgeV2.sol
@@ -211,7 +211,7 @@ contract FastBridgeV2 is Admin, IFastBridgeV2, IFastBridgeV2Errors {
                 deadline: params.deadline,
                 nonce: senderNonces[params.sender]++, // increment nonce on every bridge
                 exclusivityRelayer: paramsV2.quoteRelayer,
-                // We checked exclusivityEndTime to be in range (0 .. params.deadline] above, so can safely cast
+                // We checked exclusivityEndTime to be in range (0 .. params.deadline) above, so can safely cast
                 exclusivityEndTime: uint256(exclusivityEndTime),
                 callValue: paramsV2.callValue,
                 callParams: paramsV2.callParams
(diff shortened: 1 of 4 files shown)


packages/contracts-rfq/contracts/FastBridgeV2.sol
```diff
@@ -444,7 +447,7 @@ contract FastBridgeV2 is Admin, IFastBridgeV2, IFastBridgeV2Errors {
             revert NativeTokenCallValueNotSupported();
         }
         // exclusivityEndTime must be in range (0 .. params.deadline]
```
ChiTimesChi, line 449: Let's update the comment here to reflect that the range now includes zero.

packages/contracts-rfq/contracts/FastBridgeV2.sol
```diff
@@ -177,7 +177,10 @@ contract FastBridgeV2 is Admin, IFastBridgeV2, IFastBridgeV2Errors {
 
     /// @inheritdoc IFastBridgeV2
     function bridge(BridgeParams memory params, BridgeParamsV2 memory paramsV2) public payable {
-        int256 exclusivityEndTime = int256(block.timestamp) + paramsV2.quoteExclusivitySeconds;
+        int256 exclusivityEndTime = paramsV2.quoteRelayer != address(0)
+            // prettier-ignore
+            ? int256(block.timestamp) + paramsV2.quoteExclusivitySeconds
+            : int256(0);
```
ChiTimesChi, line 183: I think this might a bit more clear without adding a gas overhead:
```suggestion
        int256 exclusivityEndTime = 0;
        if (paramsV2.quoteRelayer != address(0)) {
            exclusivityEndTime = int256(block.timestamp) + paramsV2.quoteExclusivitySeconds;
        }
```

ChiTimesChi: Also adding any kind of generic comment wrt `exclusivityEndTime` logic here would be good for easier understanding in the future (both for us and external reviewers).

packages/contracts-rfq/test/FastBridgeV2.GasBench.Src.t.sol
```diff
@@ -50,13 +50,13 @@ contract FastBridgeV2GasBenchmarkSrcTest is FastBridgeV2SrcBaseTest {
 
     function createFixturesV2() public virtual override {
```
ChiTimesChi, line 51: This function is now basically a no-op after the parent's `createFixturesV2` is called (the default values for these structs residing in storage is zero). So you might as well remove the `createFixturesV2` definition in this particular test contract (no additional stuff is actually needed on top of the inherited fixture creation).

ChiTimesChi (changes requested):
LGTM in terms of functionality and the tests.
Requesting a few changes: a stale comment needs to be updated + a few styling nits.

packages/contracts-rfq/contracts/FastBridgeV2.sol
parodime: i actually thought it was incorrect before (0 was not technically in range before, but was specified in the comment) - but now happens to be accurate & left it alone.

ChiTimesChi: @parodime I meant purely in terms of the exclusive/inclusive range notation: `(0 .. X]` vs `[0 .. X]`

parodime: i'm blind - thanks

PR #3352, opened 2024-10-29
feat(contracts-rfq): Token Zap [SLT-389]

ChiTimesChi:
# Description
Minimal implementation of a contract that facilitates the generic Zap action for ERC20 tokens or native gas tokens (ETH/BNB/AVAX/others).

This enables **atomic** workflows like:
- Bridge funds through `FastBridgeV2`.
- On destination chain deposit the proceeds into some DeFi protocol on behalf of the original user.

> The challenge here is that we don't know how much tokens will we receive on destination chain in advance, so we can't form the **exact** calldata for the destination deposit call during the quoting phase. Therefore a hot-swapping mechanism was used, where a placeholder value is used for initial encoding, which is replaced in place by the actual bridged amount on the destination chain.

A **single** permisionless `TokenZap` contract could be used for **any** DeFi protocol and **any** FastBridge (V2+) deployment, assuming:
- The DeFi protocol supports router-like deposits on behalf of the user.
- The deposit is performed using a single function interaction (plus the token approval if needed).

# Full Workflow

Let's suppose Alice wants to deposit USDC into a Vault contract on Arbitrum. Alice has 100 USDC to start with on Optimism. The Arbitrum Vault exposes a following function to make the deposits: `depositFunds(address token, uint256 amount, address onBehalfOf)`. Following steps need to be done (abstracted away from the Alice by the FE):
1. ABI-encode calldata of the deposit call: `payload = abi.encodeCall(vault.depositFunds, (arbUSDC, 0, alice))`
   > **Note:** we don't know the destination amount yet, so we use placeholder zero value instead
2. Determine the position of the encoded `amount` within `payload`: `amountPosition = 4 + 32 * 1 = 36`
   > **Note:** this is usually 4 (function selector), plus extra 32 per every argument before `amount` in the list of parameters (regardless of the argument type). If the amount is not encoded within `payload` at all, any value greater or equal than payload length could be used in step 3.
3. Generate parameters for the Zap:
   - `TokenZap` exposes an off-chain getter: `zapData = tokenZap.encodeZapData(arbVault, payload, amountPosition)`
   - `zapNative = 0` - this should be changed only when a Zap with ERC20 requires an additional native payment (e.g. token bridged via native bridge)
4. Request a quote specifying the following parameters:
   ```yaml
    origin_chain: 10
    origin_token: "0xOP_USDC"
    origin_amount: 100 * 10**6
    origin_sender: "0xALICE"
    dest_chain: 42141
    dest_token: "0xARB_USDC"
    dest_recipient: "0xARB_TOKEN_ZAP"
    zap_data: zapData
    zapNative: 0
   ```
5. Suppose one of the Relayers responded that they are ready to relay such transaction, giving 99.9 USDC on Arbitrum.
6. Alices accepts a quote and initiates a transaction on Optimism using the parameters from step (4) and adding `dest_amount` from step (5).
7. Relayer completes the fill on Arbitrum:
   - They supply 99.9 USDC, which are transferred to TokenZap.
   - Then a Zap is performed using the Zap Data: TokenZap performs a 99.9 USDC deposit on behalf of Alice
      > **Note:** initially encoded placeholder zero value is replaced by the actual bridged amount by `TokenZap` automatically.
8. As a result, Relayer later receives 100 USDC on Optimism, while Aice has 99.9 USDC deposited on Optimism into the Vault.

## Summary by CodeRabbit

- **New Features**
	- Introduced `ZapDataV1` library for efficient data management and validation.
	- Launched `TokenZapV1` contract enabling token "Zap" actions with secure transfers.
	- Added integration tests for `FastBridgeV2` functionality related to token bridging.
	- Introduced `VaultManyArguments` mock contract for enhanced testing scenarios.

- **Bug Fixes**
	- Enhanced error handling in `TokenZapV1` for incorrect amounts and payload lengths.

- **Tests**
	- Comprehensive test suite for `TokenZapV1` and `ZapDataV1` ensuring functionality and error handling.


parodime (approved):
i added some test assumptions that were failing on my end, otherwise LGTM

ChiTimesChi:
@coderabbitai review

trajan0x (approved):
LGTM

PR #3372, opened 2024-11-06
activeQuote depositAmount [SLT-464]

parodime:
ActiveQuoting needs to consider the deposit amount when preparing an Active RFQ response and utilize it to calculate a corresponding dest amount.

## Summary by CodeRabbit

- **New Features**
	- Enhanced quote generation process to include specific deposit amounts.
	- New field for exact deposit amounts added to quote input structure.
	- Added a test for validating quote calculations based on exact deposit amounts.

- **Bug Fixes**
	- Improved error handling during deposit amount parsing and quote generation.
	- Updated error messages for invalid deposit amounts to enhance clarity.

- **Documentation**
	- Updated API documentation to reflect the change from `origin_amount` to `origin_amount_exact` for clarity and precision in data representation.


services/rfq/relayer/quoter/quoter.go
```diff
@@ -540,6 +546,7 @@ func (m *Manager) generateQuotes(parentCtx context.Context, chainID int, address
 						OriginBalance:   originBalance,
 						DestBalance:     balance,
 						DestRFQAddr:     destRFQAddr.Hex(),
+						DepositAmount:   nil,
```
ChiTimesChi, line 549: Minor nit: let's add a small comment explaining why this is `nil`.

services/rfq/relayer/quoter/quoter.go
```diff
@@ -764,6 +772,20 @@ func (m *Manager) getOriginAmount(parentCtx context.Context, input QuoteInput) (
...
+	if input.DepositAmount != nil {
+		if quoteAmount.Cmp(input.DepositAmount) >= 0 {
+			quoteAmount = input.DepositAmount
+		} else {
+			span.AddEvent("quote amount insufficient to cover deposit amount", trace.WithAttributes(
+				attribute.String("quote_amount", quoteAmount.String()),
+				attribute.String("deposit_amount", input.DepositAmount.String()),
+			))
+			quoteAmount = big.NewInt(0)
+		}
```
ChiTimesChi, line 786: Let's add some cases for this in `TestGetOriginAmount`

ChiTimesChi (changes requested):
That's a clever way to implement it, nice work!

Let's get the tests done - then should be good to merge

services/rfq/relayer/quoter/quoter_test.go
```diff
@@ -202,6 +202,248 @@ func (s *QuoterSuite) TestIsProfitable() {
 	s.False(s.manager.IsProfitable(s.GetTestContext(), quote))
 }
 
+func (s *QuoterSuite) TestGetOriginAmountActiveQuotes() {
```
dwasse, line 205: May be easier to add some test cases with the DepositAmount to the existing TestGetOriginAmount() test, but up to you

services/rfq/relayer/quoter/quoter.go
```diff
@@ -364,13 +364,19 @@ func (m *Manager) generateActiveRFQ(ctx context.Context, msg *model.ActiveRFQMes
 	}
 	span.SetAttributes(attribute.String("request_id", rfqRequest.RequestID))
 
+	depositAmount, ok := new(big.Int).SetString(rfqRequest.Data.OriginAmount, 10)
```
dwasse, line 367: Could we keep this as originAmount instead of depositAmount for consistency?

To be clear I don't really have preference between originAmount and depositAmount but I do think it's cleaner to keep the naming consistent across the codebase. If you do prefer depositAmount we can run with that, but we should change it everywhere

services/rfq/relayer/quoter/quoter_test.go
parodime: i originally did that, but the function was huge & mostly cloned from the existing ones. felt cleaner to separate

services/rfq/relayer/quoter/quoter.go
parodime: yeah youre right - actually for even more clarity i'd like to go with "originAmountExact" which is also consistent with swap interfaces

dwasse (approved):

PR #3382, opened 2024-11-11
feat(contracts-rfq): rework permisionless cancellation [SLT-489]

ChiTimesChi:
**Description**
- Users can permisionlessly cancel their orders after they expire, and additional `cancelDelay` passes.
- `cancelDelay` is configurable by the governance
  - Default value after contract deployment: 1 day
  - Can not be set lower than 1 hour.

Misc additional deprecations that became possible by dropping old `Admin` template:
- `chainGasAmount` and `setChainGasAmount` are deprecated, as they are replaced by the "zap native" V2 feature.
- `setProtocolFeeRate` uses a custom error instead of a string error message.

## Summary by CodeRabbit

- **New Features**
	- Introduced the `AdminV2` smart contract for enhanced access control and fee management.
	- Added a `cancel` function to replace the previous refund mechanism in `FastBridgeV2`, allowing cancellation of transactions.
	- Implemented new administrative functions for setting cancellation delays and protocol fee rates.
	- Enhanced the `IFastBridgeV2` interface with new transaction management capabilities.

- **Bug Fixes**
	- Improved error handling with custom errors for cancellation delays and fee rates.

- **Documentation**
	- Updated comments and documentation to reflect changes in functionality and naming conventions.

- **Tests**
	- Refactored tests to align with the new cancellation logic and updated function names.

### DIFF
--- a/packages/contracts-rfq/contracts/AdminV2.sol
+++ b/packages/contracts-rfq/contracts/AdminV2.sol
@@ -42,10 +42,16 @@ contract AdminV2 is AccessControlEnumerable, IAdminV2, IAdminV2Errors {
         _setCancelDelay(DEFAULT_CANCEL_DELAY);
     }
 
+    /// @notice Allows the contract governor to set the cancel delay. The cancel delay is the time after the transaction
+    /// deadline after which it can be permissionlessly cancelled, if it hasn't been proven by any of the Relayers.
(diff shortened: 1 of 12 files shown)


parodime (approved):
LGTM but i think since we're relabeling the roles anyway we should switch RELAYER_ROLE to PROVER_ROLE

parodime:
also retaining RELAYER_ROLE so we can continue to use that for offchain permissions such as quoting. It is not used within the contract any longer

PR #3391, opened 2024-11-13
fix(rfq-relayer): respond with failure if quote amount is zero [SLT-459]

dwasse:
## Summary by CodeRabbit

- **New Features**
	- Enhanced quote request handling for large origin amounts, improving fallback mechanisms when no active quotes are available.
- **Bug Fixes**
	- Improved error handling and response generation for quote requests, ensuring accurate feedback when no quotes are found.
- **Refactor**
	- Streamlined response construction and logic for determining quote types, enhancing code readability and modularity.

### DIFF
--- a/services/rfq/api/rest/server.go
+++ b/services/rfq/api/rest/server.go
@@ -570,7 +570,7 @@ func getQuoteResponse(ctx context.Context, quote *model.QuoteData, quoteType str
 			destAmount = amt
 		}
 	}
-	if quote == nil || destAmount.Cmp(big.NewInt(0)) <= 0 {
+	if destAmount.Sign() <= 0 {
 		span.AddEvent("no quotes found")
 		resp = model.PutRFQResponse{
 			Success: false,
--- a/services/rfq/api/rest/server.go
+++ b/services/rfq/api/rest/server.go
@@ -548,8 +548,10 @@ func (r *QuoterAPIServer) PutRFQRequest(c *gin.Context) {
 		span.SetAttributes(attribute.String("passive_quote_dest_amount", *passiveQuote.DestAmount))
 	}
 	quote := getBestQuote(activeQuote, passiveQuote)
-	quoteType := quoteTypeActive
-	if activeQuote == nil {
+	var quoteType string
+	if quote == activeQuote {
+		quoteType = quoteTypeActive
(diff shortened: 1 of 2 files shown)


services/rfq/api/rest/server.go
```diff
@@ -547,20 +548,33 @@ func (r *QuoterAPIServer) PutRFQRequest(c *gin.Context) {
 		span.SetAttributes(attribute.String("passive_quote_dest_amount", *passiveQuote.DestAmount))
 	}
 	quote := getBestQuote(activeQuote, passiveQuote)
+	quoteType := quoteTypeActive
+	if activeQuote == nil {
+		quoteType = quoteTypePassive
+	}
```
ChiTimesChi, line 554: This assumes that the active quote would always be the result of `getBestQuote`, which is a very big assumption. We should derive the quote type at the same time we choose the best quote imo.

services/rfq/api/rest/server.go
```diff
@@ -547,20 +548,33 @@ func (r *QuoterAPIServer) PutRFQRequest(c *gin.Context) {
...
-	var resp model.PutRFQResponse
-	if quote == nil {
+	destAmount := big.NewInt(0)
+	if quote != nil && quote.DestAmount != nil {
+		amt, ok := destAmount.SetString(*quote.DestAmount, 10)
+		if ok {
+			destAmount = amt
+		}
+	}
+	if quote == nil || destAmount.Cmp(big.NewInt(0)) <= 0 {
```
ChiTimesChi, line 571: Two small nits here:
1. `quote == nil` is redundant here, as it this condition is caught by `destAmount` being zero.
2. Why don't we use https://pkg.go.dev/math/big#Int.Sign for comparisons to zero?

ChiTimesChi (changes requested):
Requesting a change for `quoteType`. Plus minor nits (up to you).

services/rfq/api/rest/server.go
dwasse: Good idea- we use `getBestQuote()` in several places, so instead of modifying that func I think this would work:

```
	var quoteType string
	if quote == activeQuote {
		quoteType = quoteTypeActive
	} else if quote == passiveQuote {
		quoteType = quoteTypePassive
	}
```

PR #3399, opened 2024-11-18
feat(rfq-api): check both relayer role and quoter role

dwasse:
## Summary by CodeRabbit

- **New Features**
	- Enhanced role management for relayers, including the introduction of a new `quoterRole`.
	- Updated role verification process to allow for more flexible checks based on provided roles.

- **Bug Fixes**
	- Improved error messages for role checks to provide more context in case of failures.

- **Documentation**
	- Adjusted comments for clarity and consistency regarding role-checking logic.


services/rfq/api/rest/server.go
```diff
@@ -360,13 +360,22 @@ func (r *QuoterAPIServer) checkRoleParallel(c *gin.Context, destChainID uint32)
...
+	relayerRole := crypto.Keccak256Hash([]byte("RELAYER_ROLE"))
 	g.Go(func() error {
-		v1Addr, v1Err = r.checkRole(c, destChainID, true)
+		v1Addr, v1Err = r.checkRole(c, destChainID, true, quoterRole)
+		return v1Err
+	})
+	g.Go(func() error {
+		v1Addr, v1Err = r.checkRole(c, destChainID, true, relayerRole)
 		return v1Err
 	})
```
ChiTimesChi, line 372: The cache is contract-specific, but not binded to a role. So what likely to happen here is that we check for two different roles in the same contract, one check will result in the actual RPC call, while the second one will just reuse the cached value. Since we don't really need to check multiple roles in the same contract, I suggest to leave the role cache as is. Instead, let's only check for `RELAYER_ROLE` in V1 and for `QUOTER_ROLE` in V2. This also helps with the future cleanup once the V1 is fully deprecated.

ChiTimesChi: This could also lead to race conditions between two go routines

ChiTimesChi (changes requested):
Highlighting a race condition for checking roles

PR #3487, opened 2025-01-09
reduced origin min gas balance when quoting a route w/ origin gas token

parodime:
as origin chain gets critically low on gas, this change should discourage the relayer from quoting most routes involving that chain *except* those that will replenish the deficit gas

## Summary by CodeRabbit

- **New Features**
  - Enhanced gas sufficiency checks with a flexible threshold multiplier
  - Improved handling of native gas tokens in quote generation

- **Bug Fixes**
  - Refined gas calculation logic to better support route quoting
  - Added validation for threshold multiplier range

- **Improvements**
  - More nuanced approach to gas management across different token types
  - Streamlined error messaging for gas-related checks

### DIFF
--- a/services/rfq/relayer/inventory/manager.go
+++ b/services/rfq/relayer/inventory/manager.go
@@ -48,6 +48,8 @@ type Manager interface {
 	ApproveAllTokens(ctx context.Context) error
 	// HasSufficientGas checks if there is sufficient gas for a given route.
 	HasSufficientGas(ctx context.Context, chainID int, gasValue *big.Int) (bool, error)
+	// HasSufficientGasWithMult checks if there is sufficient gas for a given route with an optional threshold multiplier applied.
+	HasSufficientGasWithMult(ctx context.Context, chainID int, gasValue *big.Int, thresholdMultiplier *float64) (bool, error)
 	// Rebalance attempts any rebalances that could be executed across all supported tokens and chains.
 	Rebalance(ctx context.Context) error
 	// GetTokenMetadata gets the metadata for a token.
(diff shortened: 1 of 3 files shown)


services/rfq/relayer/quoter/quoter.go
```diff
@@ -772,7 +772,17 @@ func (m *Manager) getOriginAmount(parentCtx context.Context, input QuoteInput) (
 	// First, check if we have enough gas to complete the a bridge for this route
 	// If not, set the quote amount to zero to make sure a stale quote won't be used
 	// TODO: handle in-flight gas; for now we can set a high min_gas_token
-	sufficentGasOrigin, err := m.inventoryManager.HasSufficientGas(ctx, input.OriginChainID, nil)
+
+	// if the origin token is native gas, require less minimum gas on the origin chain.
+	// as origin gas gets critically low, this will discourage the relayer from quoting most routes *except* those that will replenish the deficit gas.
+	var sufficentGasOrigin bool
+	if input.OriginTokenAddr.Hex() == "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee" {
```
dwasse, line 779: Can use `util.IsGasToken()` here instead of manual check

services/rfq/relayer/quoter/quoter.go
```diff
@@ -772,7 +772,17 @@ func (m *Manager) getOriginAmount(parentCtx context.Context, input QuoteInput) (
...
+		multiplier := 0.65
```
dwasse, line 780: I would say this should probably be configurable, and we can default to 0.65 or some other sensible value

services/rfq/relayer/inventory/manager.go
```diff
@@ -48,6 +48,8 @@ type Manager interface {
 	ApproveAllTokens(ctx context.Context) error
 	// HasSufficientGas checks if there is sufficient gas for a given route.
 	HasSufficientGas(ctx context.Context, chainID int, gasValue *big.Int) (bool, error)
+	// HasSufficientGasWithMult checks if there is sufficient gas for a given route with an optional threshold multiplier applied.
+	HasSufficientGasWithMult(ctx context.Context, chainID int, gasValue *big.Int, thresholdMultiplier *float64) (bool, error)
```
dwasse, line 52: My suggestion here would be to add the `thresholdMultiplier` param to `HasSufficientGas` and remove `HasSufficientGasWithMult`- since you already have `thresholdMultiplier` as a pointer, it serves as optional param. This way we can keep the interface simple

PR #3533, opened 2025-02-24
fix(sdk-router): fixed fee calculation [SYN-49]

ChiTimesChi:
**Description**
This PR changes the way that the flat fixed fee is applied for the RFQ quote calculation. It was previously applied to the origin amount, as if the fixed fee is denominated in the origin asset,

In reality, the fixed fee is denominated in the destination asset, and is applied only after the origin/destination price is applied.

## Summary by CodeRabbit

- **New Features**
  - Enhanced processing of transaction quotes with improved validation and fee calculation.
- **Bug Fixes**
  - Adjusted conditions to correctly handle cases where transaction amounts exceed limits or fall below fee thresholds.
- **Tests**
  - Introduced new utility functions for comparing `BigNumber` values and refined test scenarios for improved clarity and consistency.

### DIFF
--- a/packages/sdk-router/src/rfq/quote.test.ts
+++ b/packages/sdk-router/src/rfq/quote.test.ts
@@ -1,4 +1,5 @@
 import { BigNumber, parseFixed } from '@ethersproject/bignumber'
+import { Zero } from '@ethersproject/constants'
 
 import {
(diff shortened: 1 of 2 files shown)


parodime (approved):
LGTM

PR #3563, opened 2025-03-05
feat: Swap Engines [SYN-29] [SYN-35]

ChiTimesChi:
**Description**

**Additional context**

**Metadata**
- Fixes #[Link to Issue]


packages/sdk-router/src/operations/swap.ts
```diff
@@ -10,7 +11,93 @@ import {
...
+): Promise<SwapQuoteV2> {
+  const input: RouteInput = {
+    chainId,
+    tokenIn: handleNativeToken(tokenIn),
+    tokenOut: handleNativeToken(tokenOut),
+    amountIn,
+    msgSender: this.swapEngineSet.getTokenZap(chainId),
+    finalRecipient: {
+      entity: to ? RecipientEntity.User : RecipientEntity.UserSimulated,
+      address: to || USER_SIMULATED_ADDRESS,
```
abtestingalpha, line 69: I see `USER_SIMULATED_ADDRESS` defined as `0xFAcefaCEFACefACeFaCefacEFaCeFACEFAceFAcE`

Is there a risk where if `to` is missing, then the `finalRecipient` would get defined as the above address? If not, do we need it here?

abtestingalpha (approved):

packages/sdk-router/src/operations/swap.ts
ChiTimesChi: @abtestingalpha we're using this pattern to enable getting quotes without the connected wallet (that is used as the final recipient). The safeguard is that we only return the router address and the quote amount with no way to construct the transaction data if `to` is missing: https://github.com/synapsecns/sanguine/blob/b26ba3a0bf2a0e6c1346578edfda047efa110c51/packages/sdk-router/src/operations/swap.ts#L86-L99

Of course, the SDK consumer can be passing a random address as `to`, but that would be their responsibility at that point.

PR #3592, opened 2025-03-12
Ft/hype rfq price feeds

parodime:
PRICE FEED CHANGES
#####################

Add ability to obtain USD prices for assets from Pyth and KuCoin in addition to CoinGecko

Add a configurable "Verification" price concept which the "Primary" price can be compared against as a safety check

PRICE CACHING CHANGES
######################
Remove both token price caching methods used previously:

tokenPriceCache on fee_pricer
and
globalCache on fetcher

Replace with a new priceCache in the fetcher.

New cache will allow token-level TTL settings (eg: we can keep USDC's price cache alive much longer than other assets)

It also handles concurrency - where if 100 requests for a price occur async at the same moment, only one of them will actually perform the fetch&cache -- the others will wait briefly for the value to finish caching & then proceed with it.


services/rfq/relayer/pricer/fetcher.go
```diff
@@ -25,133 +29,239 @@ import (
...
 }
 
-var coingeckoIDLookup = map[string]string{
-	"USDC": "usd-coin",
-	"ETH":  "ethereum",
-	"BERA": "berachain-bera",
-	"HYPE": "hyperliquid",
+type PriceSourceDetail struct {
+	Source     string
+	ExternalId string
```
ChiTimesChi, line 74: `ExternalID` ? according to the lint message

services/rfq/relayer/pricer/fetcher.go
```diff
@@ -25,133 +29,239 @@ import (
...
-	price      float64
-	source     string
-	timestamp  time.Time
-	timeToLive time.Duration
-}
-
 var (
-	globalCache = make(map[string]cachedPrice)
-	cacheMu     sync.Mutex
+	priceCache = ttlcache.New[string, float64]()
```
ChiTimesChi, line 47: Need `ttlcache.WithDisableTouchOnHit[string, float64]()`  I believe

services/rfq/relayer/pricer/fetcher.go
```diff
@@ -25,133 +29,239 @@ import (
...
+
+	// Add a separate RWMutex for tokenIDLookup
+	tokenIDMu sync.RWMutex
 
-	// Add a separate RWMutex for coingeckoIDLookup
-	coingeckoMu sync.RWMutex
+	inFlightRequests = make(map[string]*sync.Cond)
+
+	inFlightRequestsMu sync.Mutex
```
ChiTimesChi, line 54: Claude suggested using `singleflight.Group` for this. Will look something like this in `GetPrice`. Does this look good @trajan0x ?

```go
	// Use singleflight to deduplicate concurrent requests for the same token
	// This ensures only one external API call happens even with concurrent requests
	result, err, shared := priceFetchGroup.Do(token, func() (interface{}, error) {
		// Perform the actual external API request - this is the only goroutine doing the external call
		return c.fetchPriceExternal(ctx, token, tokenConfig)
	})

	if err != nil {
		return 0, err
	}

	span.SetAttributes(attribute.Bool("used_cached_price", shared))

	return result.(float64), nil
```

services/rfq/relayer/pricer/fetcher.go
```diff
@@ -226,62 +389,160 @@ func (c *CoingeckoPriceFetcherImpl) GetPrice(ctx context.Context, token string)
...
+	if currentTime-resp.Parsed[0].Price.PublishTime > 5*60 {
+		return price, fmt.Errorf("err stale price (%d)", resp.Parsed[0].Price.PublishTime)
+	}
 
-	if secondarySource != "None" {
-		// obtain secondary price for the asset
-		secondaryPrice := 5.0 //, err := c.GetSecondaryPrice(ctx, token)
+	if len(resp.Parsed) == 0 {
+		return price, fmt.Errorf("err no price data found")
+	}
```
ChiTimesChi, line 422: Duplicate check few lines above

services/rfq/relayer/pricer/fetcher.go
```diff
@@ -226,62 +389,160 @@ func (c *CoingeckoPriceFetcherImpl) GetPrice(ctx context.Context, token string)
...
+	if err != nil {
+		return price, fmt.Errorf("err sending request: %w", err)
+	}
+	if r.StatusCode != http.StatusOK {
+		return price, fmt.Errorf("err response status: %v", r.Status)
+	}
+	defer func() {
+		closeErr := r.Body.Close()
+		if closeErr != nil {
 		}
```
ChiTimesChi, line 456: This looks a bit sus

services/rfq/relayer/pricer/fetcher.go
```diff
@@ -226,62 +389,160 @@ func (c *CoingeckoPriceFetcherImpl) GetPrice(ctx context.Context, token string)
...
+		return price, fmt.Errorf("err no price data found for token: %s", kucoinTokenId)
 	}
-	cacheMu.Unlock()
 
 	return price, nil
 }
 
-func (c *CoingeckoPriceFetcherImpl) GetSecondaryPrice(ctx context.Context, token string) (price float64, err error) {
-	return 1, nil
+// getPriceFromCoinGecko fetches the price of a token from CoinGecko.
```
ChiTimesChi, line 492: Need to update the func name

services/rfq/relayer/pricer/fetcher.go
ChiTimesChi: Also don't we need to do priceCache.Start()?

ChiTimesChi (changes requested):

services/rfq/relayer/pricer/fetcher.go
parodime: yes i totally missed the disable touch, thanks

also apparently Start() not needed. AI says its automatic w/ .New -- also seems to run fine w/ local testing

services/rfq/relayer/pricer/fetcher.go
parodime: singleflight was suggested when I checked w/ AI on this originally, but I also asked about a solution w/ no additional lib dependency just to see.

It seemed to be simple enough & went with the latter - also seems to work fine in local tests, for what that is worth


I'm ok with switching if we want to - will leave it as-is until @trajan0x can weigh in

services/rfq/relayer/pricer/fetcher.go
parodime: just relabeled to SourceTokenID

ChiTimesChi (approved):

PR #4030, opened 2026-04-30
fix(sdk-router): enable SBA routes to Ethereum

zeroxzerollc:
**Description**
Enable SynapseBridgeAdapter (SBA) quote discovery for routes whose destination chain is Ethereum.

SBA already has Ethereum chain metadata and artifact-backed token mappings, including Arbitrum gOHM -> Ethereum gOHM. However, `SynapseBridgeAdapterModuleSet.getBridgeTokenCandidates()` currently rejects Ethereum as a destination before consulting the artifact. This causes Bridge V2 quote discovery to skip the native/SBA route and fall through to other modules such as Relay.

This change adds `SupportedChainId.ETH` to the SBA destination enablement set and updates the module-set test to assert that Arbitrum gOHM -> Ethereum gOHM is returned as an artifact-backed candidate.

**Additional context**
This is intentionally narrow in code surface: it does not change the supported token artifact, adapter metadata, route finalization, or transaction generation. It only allows destination-Ethereum routes that are already present in the committed SBA artifact and have metadata for both chains.

Context observed while testing the current bridge UI: Arbitrum gOHM -> Ethereum gOHM receives a Relay quote that swaps through thin Arbitrum liquidity rather than using the bridge-backed gOHM route. For a 1 gOHM quote, the Relay path showed very high price impact, consistent with swap routing rather than native bridge redemption.

**Verification**
- `git diff --check` passed.
- CodeRabbit review reported no actionable comments.
- Local targeted test command was attempted but blocked by dependency access in this environment. `packages/sdk-router` could not resolve all repo dependencies locally because several `@synapsecns/*` packages appear to require upstream/private registry access. The added test assertion is included for upstream CI validation.

### DIFF
--- a/packages/sdk-router/src/sba/synapseBridgeAdapterModuleSet.test.ts
+++ b/packages/sdk-router/src/sba/synapseBridgeAdapterModuleSet.test.ts
@@ -19,9 +19,11 @@ import { SynapseBridgeAdapterModuleSet } from './synapseBridgeAdapterModuleSet'
 const ETH_AGEUR = '0x1a7e4e63778B4f12a199C062f3eFdD288afCBce8'
 const OP_AGEUR = '0xa0554607e477cdC9d0EE2A6b087F4b2DC2815C22'
 const ETH_DOG = '0xBAac2B4491727D78D2b78815144570b9f2Fe8899'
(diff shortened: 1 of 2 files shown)


zeroxzerollc:
@coderabbitai review

